PR #14, opened 2018-05-18
[suggestion] Adding a demo section

ThomasDelteil:
Thanks for the great library, I used it to create a demo website to showcase how you can build a Visual Search engine using MXNet and HNSW: https://thomasdelteil.github.io/VisualSearch_MXNet/

I was thinking it could be a good idea to add a demo section in order to point to the website to provide other users with ideas and maybe inspiration. It would be great to also see how other users are using this library if there are any other demos out there.

If such a section is not appropriate feel free to close the PR, I really won't be offended 😄


yurymalkov:
Cool! Thanks!

PR #16, opened 2018-05-29
remove using namespace std

wangty6:
`using namespace` in header file can pollute the global namespace, and is very dangerous.


yurymalkov:
Cool! Thanks!

PR #25, opened 2018-06-07
hnswlib/hnswalg.h: increase linked list element size by 1

piem:
This prevents an `Invalid read of size 4` in the `searchBaseLayer`
(hnswalg.h:194).

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -612,8 +612,8 @@ namespace hnswlib {
 
 
             if (curlevel) {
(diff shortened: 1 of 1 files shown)


yurymalkov:
Cool! Thanks! 
I am not sure if it has any effect though.

piem:
Hi @yurymalkov and thank you for both merges.

I don't think it has any side effect, but valgrind was rightly complaining about uninitialized memory being accessed in that line, since the `+ 1` would reach outside of the array bounds.

Issue #15, opened 2018-05-29
Cosine similarity is abnormally slow


wangty6:
Compared to l2 and ip, cosine is slow abnormally and the index time grows exponentially after the point is more than 10000.
data size=14000
index           query
4021134      5241054     l2
5447251      7004991     ip
23386483    65030850   cosine
time unit=1us

After some experiment, I found that the technology in `denorm` branch solved the problem, hope this can be merge into the master branch
`norm=1.0/sqrtf(norm+1e-30f);`

And I wonder why this algorithm is so sensitive to this situation, thanks very much.

yurymalkov:
Hi @wangty6 ,
Thanks!

This is probably due to zeros in the dataset, which lead to zero division during vector normalization.
The fix can be merged right now. However, the distances for such vectors are fundamentally undefined. Probably, a there should be an exception or, at least, user warning.

searchivarius:
@yurymalkov I think the suggestion from this recent issue makes most sense: https://github.com/nmslib/nmslib/issues/321
if you have zero norms, set the distance to maximum possible value.

yurymalkov:
@searchivarius Maybe it is better just not to add them to the dataset and print user warnings? That seems to be a strong sign that something is wrong with the data.

searchivarius:
@yurymalkov this is a better option, b/c it has no runtime penalty. But, maybe, one should throw an exception instead? One small difficulty though is that we use the same data loader for different spaces. So, the load should be somehow made aware of this issue.

yurymalkov:
Throwing exception will prevent running full ann benchmarks, which has zero vectors in some datasets...
Yeah, probably loader is the best place to skip them(together with labels).

PR #49, opened 2018-08-10
Add denormalization coefficient

yurymalkov:
Should be a temporary solution.


yurymalkov:
Solves #15, but more general solution is needed.

PR #53, opened 2018-08-22
Dupcode

ironyoung:


ironyoung:
rename bruto->brute & delete duplicated codes

yurymalkov:
@ironyoung Thank you for the fix!

PR #57, opened 2018-09-04
Add get data by label

ironyoung:


hnswlib/hnswalg.h
```diff
@@ -616,6 +617,20 @@ namespace hnswlib {
             return;
         }
 
+        std::vector<dist_t> getDataByLabel(labeltype label)
```
yurymalkov: Type dist_t is for distances (e.g. result of the distance function), while the data vectors may have different type, so probably, a separate template type for this method should be better.

python_bindings/bindings.cpp
```diff
@@ -212,13 +213,26 @@ class Index {
                 });
             };
             cur_l+=rows;
-
-
         }
+    }
 
+    std::vector<std::vector<dist_t> > GetDataReturnNumpy(py::object ids_ = py::none()) {
```
yurymalkov: This method returns a vector object which is later converted to a plain python list. Does not seem be an issue, since it is not performance critical, but the name of the method is a bit confusing (GetDataReturn*Numpy*).

yurymalkov:
@ironyoung Do my comments make sense?

ironyoung:
sorry I'm too busy this week... I've updated this Pull Request, pls review it

yurymalkov:
@ironyoung Thank you! I am in a work trip right now, I'll check it as soon as I will be back.

yurymalkov:
Hi @ironyoung,
Can you move the template? If you have no time, I can handle it.

yurymalkov:
It seems that save/load functionality is broken due to ```label_lookup_``` is not being restored during the index loading.

yurymalkov:
@ironyoung I've merged your code to https://github.com/nmslib/hnsw/tree/develop; and added changes to deal with the above-mentioned issues. 
Can you please check the code is working for your application?

yurymalkov:
Merged via a different pull request

ChasenYang:
I have some questions about the getDatabylabel interface
When using Cosine similarity to calculate distance, the writing interface(python_bindings/bindings.cpp) uses normalize_vector for normalization. So if you need to get the original value after querying, should provide a denormalize_vector?

PR #66, opened 2018-10-17
Add some const qualifiers to enable usage of searchKnn in const contexts

dgenzel:
This is a trivial change to enable users to instantiate a const object and still be able to call searchKnn.


yurymalkov:
@dgenzel Thank you!

Issue #67, opened 2018-10-20
Compiler pedantry: unnecessary semi-colons


jlmelville:
If running `g++` in `-pedantic` mode, there are warnings about unnecessary semi-colons terminating the static function definitions and `namespace` declarations in `space_ip.h` and `space_l2.h`.

jlmelville:
Also, the following variable is unused: https://github.com/nmslib/hnsw/blob/dbac49b2a6d3ed9a4d22e1b5ea0903106b2c2855/hnswlib/space_ip.h#L174

PR #68, opened 2018-10-20
Remove unnecessary semi-colons

jlmelville:
A fix for the trivial issue #67. Recent changes to the [R devtools package](https://cran.r-project.org/package=devtools) defaults causes these warnings to appear when building [R bindings](https://github.com/jlmelville/rcpphnsw) to hnsw.


yurymalkov:
@jlmelville Thank you!

PR #84, opened 2018-12-20
Raising an error when trying to get data for a non-existent label

SeanU:
If you query `getDataByLabel` for a label that's not in the index, it will return a vector with un-initialized data. I believe, because of the way `std::unordered_map`'s `[]` operator works, it will also add that random vector to the map.

This change instead returns the data when that label exists, and otherwise throws an error.

I also added another Python test to verify this behavior.


yurymalkov:
@SeanU many thanks for the fix!

PR #85, opened 2018-12-22
Add preprocessor directives for non-portable code.

jlmelville:
In order to submit the [R bindings for HNSW](https://github.com/jlmelville/rcpphnsw) to CRAN, it is a requirement that C++ code build without needing non-portable flags like `-march=native`. @LTLA and I have made those changes and we were wondering if this might also be useful for those hoping to use HNSW on architectures without AVX, SSE and the like. Hence this PR.

Changes in roughly decreasing order of importance:
* Added a preprocessor token`NO_MANUAL_VECTORIZATION`. If defined (e.g. by using `-DNO_MANUAL_VECTORIZATION` in the compiler flags), then you don't need to specify `-march=native`  and AVX and SSE will not be used. If you don't define `NO_MANUAL_VECTORIZATION` then the code will build as before, i.e. anyone who doesn't care about this should be unaffected.
* There are some other tokens defined such as `USE_SSE` and `USE_AVX`, but these are intended for internal use only.
* Added some virtual destructors.
* Removed an unused variable.

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -184,15 +184,19 @@ namespace hnswlib {
                     data = (int *) (linkLists_[curNodeNum] + (layer - 1) * size_links_per_element_);
                 int size = *data;
                 tableint *datal = (tableint *) (data + 1);
+        #ifdef USE_SSE
                 _mm_prefetch((char *) (visited_array + *(data + 1)), _MM_HINT_T0);
                 _mm_prefetch((char *) (visited_array + *(data + 1) + 64), _MM_HINT_T0);
                 _mm_prefetch(getDataByInternalId(*datal), _MM_HINT_T0);
                 _mm_prefetch(getDataByInternalId(*(datal + 1)), _MM_HINT_T0);
+        #endif
 
                 for (int j = 0; j < size; j++) {
                     tableint candidate_id = *(datal + j);
+        #ifdef USE_SSE
                     _mm_prefetch((char *) (visited_array + *(datal + j + 1)), _MM_HINT_T0);
                     _mm_prefetch(getDataByInternalId(*(datal + j + 1)), _MM_HINT_T0);
+        #endif
                     if (visited_array[candidate_id] == visited_array_tag) continue;
                     visited_array[candidate_id] = visited_array_tag;
                     char *currObj1 = (getDataByInternalId(candidate_id));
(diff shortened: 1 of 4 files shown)


yurymalkov:
Hi @jlmelville ,
Thank you for the pull request! I'll perform manual speed checks on Monday.
I suspect that Travis was updated and exposed a bug in the test caused it to fail.

LTLA:
Upon thinking a bit more about point 1 above; the original HNSW library would use SSE intrinsics all the time, without any regard for whether the CPU or OS was able to support them. The new macro set-up now protects against this by checking for `__SSE__`, even when `NO_MANUAL_VECTORIZATION` is not set. This is safer as it avoids problems on architectures without SSE instructions. However, it means that users will have to explicitly set `-msse` or related flags in order to use SSE, which they didn't have to do before. Of course, if people were already setting `-march=native`, then no additional work is required.

yurymalkov:
@LTLA Ok, thanks! AFAIK there are no 64-bit x86 CPUs that don't support SSE. So ```NO_MANUAL_VECTORIZATION``` seems like a switch between x86/non-x86 platforms.

jlmelville:
Linking to @LTLA comments on reasons for the CI failure: https://github.com/jlmelville/rcpphnsw/issues/1#issuecomment-449562539

LTLA:
@yurymalkov The new internal macros (namely `USE_SSE` and `USE_AVX`) should automatically act as switches for different architectures that support different levels of intrinsics, regardless of whether `NO_MANUAL_VECTORIZATION` is set. The purpose of `NO_MANUAL_VECTORIZATION` is to forbid the use of any intrinsics to ensure standard-compliant portable code that behaves the same across architectures. This avoids inconsistent round-off errors from differently ordered summation when computing distances with and without SSE instructions; and allow binaries compiled on one machine to be distributed.

Looking at Intel's CPU history indicates that the last CPU without SSE instructions was the Pentium II (geez - I was still in elementary school when that came out!) so perhaps the assumption of SSE availability is pretty reasonable for any modern scientific computing environment. Nonetheless, I think it's generally nice to have an option to only use standard C++.

yurymalkov:
@LTLA @jlmelville I've check the commit. Everything is fine. Thank you! 

Also I've checked the performance slowdown without SSE on Ryzen - a little less than 2X for SIFT1M.

PR #94, opened 2019-03-01
Remove link to OCaml implementation.

lehy:
I am the author of the OCaml implementation. I was not able to replicate your performance results, and I suspect something is broken in my implementation. I'd prefer not to point people to it at the moment. Thanks for HNSW!


yurymalkov:
Ok. Thank you!

PR #98, opened 2019-04-22
mod: add two getters, so that we can check the consistancy of the arg…

ctero-graham:
…ument when service resumes by loading existed index file.


yurymalkov:
Cool! Thanks!

PR #101, opened 2019-05-15
support deletion

ctero-graham:
Deletion is support in two ways: markDelete and recycle.

By markDeleted, a deleted flag is set at the higher 8 bits of the link size at level 0, no extra memory needed, the drawback is that the size of the links will be limited to 24bits, however, large enough in almost cases. 

By recycle(now named recycle_in_test), all the nodes marked deletion will be kicked out, the slots they occupied will be recorded in a similar way to linked list, by adding two vars: reusable_entry / reusable_tail, and reusing the memory for label to store the next.

recycle should be used after some "markDeleted"s, for a small number of deletion, only markDeleted would be enough, it depends on the scenario how to combine with recycle.

I did some testing, it seems ok for markDeleted, but not that sure for recycle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,33 @@ Note that inner product is not a metric. An element can be closer to some other 
 
 For other spaces use the nmslib library https://github.com/nmslib/nmslib. 
 
+#### short API description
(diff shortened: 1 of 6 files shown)


vinnitu:
hm, I don't see recycle method

yurymalkov:
@vinnitu 
There is no recycle in the code.

PR #105, opened 2019-05-28
deletion by marks supported

ctero-graham:
Support deletion by adding a "deleted" mark to the elements.

The higher 8 bits of the list link size at level 0 will be used the store the mark, therefore, the link size has to be limited to the lower 24 bits, however, still large enough for almost all cases.

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -3,17 +3,9 @@
 #include "visited_list_pool.h"
 #include "hnswlib.h"
 #include <random>
-#include <iostream>
-#include <fstream>
-#include <stdio.h>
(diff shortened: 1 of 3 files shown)


yurymalkov:
There are some lines with commented code left, but overall seems good to me. 
I can merge it.

Thank you so much!

ctero-graham:
Hi @yurymalkov 
We have some projects depending on your repo, please let me know when it is merged into master, so that we can update the dependency and provide feature of deletion.

yurymalkov:
@ctero-graham Sure! 
I'll merge it by the end of the next week (few other changes are needed to be done, I am working on them right now).
Thanks again for the PR!

PR #127, opened 2019-07-24
use auto counters to avoid size_t vs. int comparisons

piem:
howdy,

here is a simple patch to avoid warnings on systems where `sizeof(size_t) != sizeof(int)`.

cheers, piem


piem:
actually, using `size_t` is required to avoid the warning. will re-submit PR,

cheers, piem

PR #128, opened 2019-07-24
setExternalLabel returns void

piem:
howdy,

This patch fixes the prototype of `setExternalLabel`, which has no return value.

cheers, piem

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -131,7 +131,7 @@ namespace hnswlib {
             return return_label;
         }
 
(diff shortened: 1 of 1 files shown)


yurymalkov:
Hi @piem,

I am not sure, why the prototype needs to return a value?

piem:
hi @yurymalkov 

currently the function prototype says it returns a `labeltype`, but its implementation doesn't return anything. this patch fixes the prototype.

another, safer option would be to return a `bool` to check that `memcpy` succeeded.

best, piem

yurymalkov:
Got it! I am a bit slow :) Thank you!

PR #129, opened 2019-07-24
use size_t counters to avoid size_t to int comparisons

piem:
howdy again,

here is a simple patch to avoid warnings on systems where sizeof(size_t) != sizeof(int).

cheers, piem


yurymalkov:
Thank you!

PR #131, opened 2019-07-30
[L2 Space] Improving performance when dimension is not a factor of 4 or 16

2ooom:
Processing 8 values at once and finishing computation by non-vectorized instructions in case dim % 8 != 0


yurymalkov:
Hi @2ooom,
Thanks for the PR!

A long time ago I've done the performance tests and it turns out that the computation of the residual does contribute to the total run time even if `dim%16==0` (no residual). So, I would like to keep the conditions of the exactness along with the others for the best performance in this case.

This probably can be done without much duplicate coding by using templates with bools in `L2SqrSIMD16Ext` (e.g. if  `dim%16==0` then set a bool to skip the tail at the compile time; if `dim%16!=0` then process the tail).

Can you please update the PR to accommodate that?

PR #132, opened 2019-08-10
Fix USE_AVX flag

tanliboy:
USE_AVX and USE_SSE flags should be defined independently.

### DIFF
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -2,9 +2,9 @@
 #ifndef NO_MANUAL_VECTORIZATION
 #ifdef __SSE__
 #define USE_SSE
+#endif
 #ifdef __AVX__
 #define USE_AVX
-#endif
 #endif
 #endif
 


tanliboy:
@jlmelville would you mind helping review?

jlmelville:
Paging @LTLA on this one: are these flag intended to be used independently?

LTLA:
No, but there's no real reason that they couldn't be.

I just figured that all processors capable of AVX would also support SSE, hence the current set-up. It probably makes more sense to have them checked independently, if there are indeed architectures that have AVX but not SSE. (Are there? I'm not an expert on that.)

That said, you will have to go and double-check that there are no SSE intrinsics being used in the AVX-marked sections of the code. For example, in:

https://github.com/nmslib/hnswlib/blob/79c74b45182363eb794c9fd19a7f099fa447a36e/hnswlib/space_l2.h#L152-L158

there is an assumption that `USE_SSE` is defined when `USE_AVX` is defined, as `L2SqrSIMD4Ext` is only defined when `USE_SSE` is defined.

tanliboy:
Thank you for the explanation, @LTLA !
I will abandon this PR.

Issue #146, opened 2019-09-14
About interface


uestc-lfs:
I'm integrating hnswlib into my own code, I have the following suggestions:

1. I think that the data passed to `addPoint` should be `const void *` .

https://github.com/nmslib/hnswlib/blob/44f20f0b2257af1b3251ac9ed976a6d6583967c0/hnswlib/hnswalg.h#L773

2. I think that `searchKnn` shuld first check if the "container" is empty, in wich case a search may result in a core dump.
https://github.com/nmslib/hnswlib/blob/44f20f0b2257af1b3251ac9ed976a6d6583967c0/hnswlib/hnswalg.h#L486

uestc-lfs:
Another suggestion: the result points are aranged in a max heap, while what people interested in is the nearest points. I would suggest using another comparator that would return a min heap.
            std::priority_queue<std::pair<dist_t, labeltype >> results;
            while (top_candidates.size() > k) {
                top_candidates.pop();
            }
            while (top_candidates.size() > 0) {
                std::pair<dist_t, tableint> rez = top_candidates.top();
                results.push(std::pair<dist_t, labeltype>(rez.first, getExternalLabel(rez.second)));
                top_candidates.pop();
            }
            return results;

yurymalkov:
@uestc-lfs Thanks for the suggestions! Seems reasonable.

uestc-lfs:
Would you like some pull requests for these suggestions?

yurymalkov:
@uestc-lfs 
Yeah, sure. That would be great!
I've already started to implement them, but was distracted by other things...

PR #153, opened 2019-09-25
Interface addpoint:change void* to const void*

uestc-lfs:
change from addPoint(void*, args...) to addPoint(const void*, args...).

The changes include the interface in bruteforce.h, and all interfaces
related to addPoint in hnswlib.

I test the code using the 1 million bigann data, the result is ok.

This is from issue #146


hnswlib/hnswlib.h
```diff
@@ -60,8 +77,8 @@ namespace hnswlib {
     template<typename dist_t>
     class AlgorithmInterface {
     public:
-        virtual void addPoint(void *datapoint, labeltype label)=0;
-        virtual std::priority_queue<std::pair<dist_t, labeltype >> searchKnn(const void *, size_t) const = 0;
+        virtual void addPoint(const void *datapoint, labeltype label)=0;
+        virtual retType<dist_t> searchKnn(const void *, size_t) const = 0;
```
yurymalkov: Changing the result of the `searchKnn` breaks backward compatibility with all C++ code...
I think adding a separate method with template comparator would be a better solution, isn't it?

uestc-lfs:
Yes,I realized that. I will provide another interface tomorrow.  Besides, will change to const void break the compatibility with python interface?

yurymalkov:
@uestc-lfs I think it is ok to change the python interface as it is an internal part of the library.

hnswlib/bruteforce.h
```diff
@@ -106,6 +109,28 @@ namespace hnswlib {
...
+            if (cur_element_count == 0) return result;
+
+            auto ret = searchKnn(query_data, k);
+
+            while (!ret.empty()) {
+                result.push_back(ret.top());
+                ret.pop();
+            }
+
+            if (result.size() > 1) {
```
yurymalkov: This additional loop can be avoided if filling the `result` from the back. This doesn't seem to be a big performance issue.

hnswlib/hnswalg.h
```diff
@@ -484,6 +484,8 @@ namespace hnswlib {
 
 
         std::priority_queue<std::pair<dist_t, tableint>> searchKnnInternal(void *query_data, int k) {
+            std::priority_queue<std::pair<dist_t, tableint  >> top_candidates;
+            if (cur_element_count == 0) return top_candidates;
```
yurymalkov:
```suggestion
            if (cur_element_count == 0) {
                  std::priority_queue<std::pair<dist_t, tableint  >> top_candidates;
                  return top_candidates;
            }
```

yurymalkov: Not sure if the compiler does not optimize this, but there is no need to create `top_candidates` if it does not branch.

hnswlib/bruteforce.h
uestc-lfs: Not quite understand. Can you show me how to do that?

yurymalkov: Something like:
```c++           
            if (cur_element_count == 0) {
                std::vector<std::pair<dist_t, labeltype>> result;
                return result;
            }
            
            auto ret = searchKnn(query_data, k);
            
            std::vector<std::pair<dist_t, labeltype>> result(ret.size);
            int i=ret.size-1;
            
            while (!ret.empty()) {
                result[i]=ret.top();
                ret.pop();
                i-=1;
            }
            return result;
```

uestc-lfs:
You assume that ret has a descending order and result has a ascending order. However, I want to return the result according to the order specified by the comparator. And I also assume that the order of ret is unknown.

2019年9月28日 上午12:43，Yury Malkov <notifications@github.com>写道：

@yurymalkov commented on this pull request.

________________________________

In hnswlib/bruteforce.h<https://github.com/nmslib/hnswlib/pull/153#discussion_r329159699>:

> @@ -106,6 +109,28 @@ namespace hnswlib {
             return topResults;
         };

+        template <typename Comp>
+        std::vector<std::pair<dist_t, labeltype>>
+        searchKnn(const void* query_data, size_t k, Comp comp) {
+            std::vector<std::pair<dist_t, labeltype>> result;
+            if (cur_element_count == 0) return result;
+
+            auto ret = searchKnn(query_data, k);
+
+            while (!ret.empty()) {
+                result.push_back(ret.top());
+                ret.pop();
+            }
+
+            if (result.size() > 1) {


Something like:

            if (cur_element_count == 0) {
                std::vector<std::pair<dist_t, labeltype>> result;
                return result;
            }

            auto ret = searchKnn(query_data, k);

            std::vector<std::pair<dist_t, labeltype>> result(ret.size);
            int i=ret.size-1;

            while (!ret.empty()) {
                result[i]=ret.top();
                ret.pop();
                i-=1;
            }
            return result;

yurymalkov:
You are right. 

However, it seems that this logic (check only for the edge elements https://github.com/nmslib/hnswlib/pull/153/files#diff-8ef7e92689e692a09f2a010888a5c7f8R126) works only for two types of comparators (strictly ascending or descending for the first element in tuple). 

This might confuse the user.  I think it might be better to directly use `std::sort`. What do you think?

uestc-lfs:
Yes, std::sort will cover all case.

I'll add it later.

2019年9月28日 上午4:56，Yury Malkov <notifications@github.com>写道：

You are right.

However, it seems that this logic (check only for the edge elements https://github.com/nmslib/hnswlib/pull/153/files#diff-8ef7e92689e692a09f2a010888a5c7f8R126) works only for two types of comparators (strictly ascending or descending for the first element in tuple).

This might confuse the user. I think it might be better to directly use std::sort. What do you think?

uestc-lfs:
@yurymalkov My email shows that there are another two comments:
image
image
But I could not see them on the pull request comments.

Why is putting `top_candidates` inside the `if` block better?

yurymalkov:
@uestc-lfs 
Thanks, will merge after the update.
I've deleted these comments, since I understood they were wrong.

yurymalkov:
@uestc-lfs Thank you!

The tests for the commit failed, but it is cause by a problem in the tests themselves.

PR #155, opened 2019-09-26
Interface addpoint

uestc-lfs:
For new interface searchKnn,
1. virtual is not used because it's a template function
2. return a `vector<pair<dist_t, labeltype>>` because it is easier to use while keeps the order


uestc-lfs:
I don't know why this becomes a new pull request..

PR #165, opened 2019-11-11
Throw exception on malloc fails

vimal-mathew:
Throw runtime_error when malloc fails


yurymalkov:
@vimal-mathew  Thank you!

PR #168, opened 2019-11-16
deep copy dist func params into AlgorithmInterface

zouyxdut:


yurymalkov:
Hi @zouyxdut,

Thank you!

I've checked the commit and it turns out it causes noticeable performance degradation (~7-8%).
This seems to be caused by the overhead of constant `.get()`. This probably by calling it once inside the constructor and using a (void *) copy, while still solving the problem with lifetime.

Can you please also redo the PR against the `develop` branch?

PR #224, opened 2020-06-27
currObj should be updated as the closest from all candidated.

### LINKED ISSUE #115: Why is entor point not updated?
https://github.com/nmslib/hnswlib/blob/68c5531caa8be522b19ac20088ffca39d1474520/hnswlib/hnswalg.h#L842-L854

This is different from the algorithm in the paper
image

I think that `currObj` should be set to `top_candidates.top().second`.

uestc-lfs:
fix #115 `currObj` shuld be updated as the nearest node in `topCandidates`

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -406,7 +406,7 @@ namespace hnswlib {
                 top_candidates.pop();
             }
 
-            tableint next_closest_entry_point = selectedNeighbors[0];
+            tableint next_closest_entry_point = selectedNeighbors.back();
 
             {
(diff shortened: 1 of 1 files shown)


uestc-lfs:
Why is this not merged, I believe it is a bug.

yurymalkov:
Good question. It seems I just forgot about it. Thanks for reminding!
I'll check it and the other PR within this week.

yurymalkov:
Thank you so much! It is a bug indeed.
Sorry it took so long. It seems I rely to much on my memory.

PR #225, opened 2020-06-27
Remove temlate interface searchKnn

uestc-lfs:
In issue #146, i add a templae interface 
https://github.com/nmslib/hnswlib/blob/3c6a84fb54a2a92dc4b7dc3e13dbae3d99892444/hnswlib/hnswlib.h#L74-L76
But it cannot be overrided. So I removed the template and make it a virtual function.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,4 +23,6 @@ endif()
 
 add_executable(test_updates examples/updates_test.cpp)
 
+add_executable(searchKnnCloserFirst_test examples/searchKnnCloserFirst_test.cpp)
+
 target_link_libraries(main sift_test) 
--- a/examples/searchKnnCloserFirst_test.cpp
+++ b/examples/searchKnnCloserFirst_test.cpp
@@ -1,0 +1,84 @@
+// This is a test file for testing the interface
+//  >>> virtual std::vector<std::pair<dist_t, labeltype>>
+//  >>>    searchKnnCloserFirst(const void* query_data, size_t k) const;
+// of class AlgorithmInterface
+
+#include "../hnswlib/hnswlib.h"
+
+#include <assert.h>
+
+#include <vector>
+#include <iostream>
+
+namespace
+{
+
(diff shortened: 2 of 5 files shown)


uestc-lfs:
Why is this pr not merged?

yurymalkov:
Thanks for the PR! And sorry for an extremely late response.
One thing it seems it breaks the bruteforce C++ code (we do not have tests for it). This can be easily fixed by moving the function to the interface (as it does not depend on the algorithm).
Secondly, I think having two methods with the same name, but different semantics is not a good move. I think renaming the function to explicitly represent the difference `knnSearchFillCloserFirst` might help with that.

uestc-lfs:
Thank you for your reply. So we should only keep the following interface
https://github.com/nmslib/hnswlib/blob/2571bdb6ef3f91d6f4c2e59178fde49055d2f980/hnswlib/hnswlib.h#L73

And add interfaces like `knnSearchFillCloserFirst`.
Am I right?

yurymalkov:
Yes. That would work!
 It would be nice if the interface would have the same semantics (input, output types)

uestc-lfs:
image
@yurymalkov which interface do you prefer. One returns a `vector` while another returns a `priority_queue`.

yurymalkov:
Hi @uestc-lfs , after some thinking I agree the vector option should be actually better as it is more flexible and can be used both ways by the user.
Probably even a single order option would be enough for practical applications (CloserFirst). Please let me know what you think.

uestc-lfs:
> Hi @uestc-lfs , after some thinking I agree the vector option should be actually better as it is more flexible and can be used both ways by the user.
> Probably even a single order option would be enough for practical applications (CloserFirst). Please let me know what you think.

I agree

uestc-lfs:
@yurymalkov I have finished my code. How do you like it?

yurymalkov:
Great! Thank you so much! And apologies that it took so long!

PR #226, opened 2020-06-28
improve performance when dim%16>=4

sjwsl:
When using SIMD16ExtResiduals and size of tail >= 4, we can use SIMD4ExtResiduals rather than non-vectorized method to calculate it.


yurymalkov:
Sorry for the delay, overloaded with other stuff. Did you test that it is actually faster than the baseline?

sjwsl:
Sorry for late reply. I did some simple tests. The effect is most obvious when dim = 28. As the percentage of the tail gradually decreases, this optimization effect will not be significant.
-------------------------
Benchmark          Time         
-------------------------
TstDim28            6.4 ns       
RefDim28            20.5 ns        
TstDim44            12.7 ns        
RefDim44            26.6 ns      
TstDim60            17,5 ns
RefDim60            33.2 ns

PR #233, opened 2020-07-14
GCC sign_compare and reorder warning flags patches

jwimberley:
In order to compile with -Werror=sign_compare and -Werror=reorder, a few small int -> size_t replacements and constructor initialization list.

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -26,7 +26,7 @@ namespace hnswlib {
         }
 
         HierarchicalNSW(SpaceInterface<dist_t> *s, size_t max_elements, size_t M = 16, size_t ef_construction = 200, size_t random_seed = 100) :
-                link_list_locks_(max_elements), element_levels_(max_elements), link_list_update_locks_(max_update_element_locks) {
+                link_list_locks_(max_elements), link_list_update_locks_(max_update_element_locks), element_levels_(max_elements) {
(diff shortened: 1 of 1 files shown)


hnswlib/hnswalg.h
```diff
@@ -26,7 +26,7 @@ namespace hnswlib {
         }
 
         HierarchicalNSW(SpaceInterface<dist_t> *s, size_t max_elements, size_t M = 16, size_t ef_construction = 200, size_t random_seed = 100) :
-                link_list_locks_(max_elements), element_levels_(max_elements), link_list_update_locks_(max_update_element_locks) {
+                link_list_locks_(max_elements), link_list_update_locks_(max_update_element_locks), element_levels_(max_elements) {
```
yurymalkov, line 29: Hm. I wonder, why the order matter here?

jwimberley: I the order only really matters pragmatically, for when a client (such as myself) needs to include hnswlib in a project that builds with the -Wall -Werror compiler flags by policy, in which case this won't compile. I doubt there's actually any problem in this particular case, and the warning is just a false positive, but https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder explains the justification for this GCC warning flag in general.

yurymalkov: I see. Thanks! Didn't know about this.

PR #241, opened 2020-08-14
Update ALGO_PARAMS.md

vinnitu:
is it correct?


yurymalkov:
@vinnitu Thanks for the PR!
I am not sure that calling dimensionality `d` instead of `dim` is better. Probably, it makes sense to directly define the parameter,

vinnitu:
image

not the same as I understand

image

yurymalkov:
Makes sense. Thanks!

PR #243, opened 2020-08-19
L2SqrI: add fallback if the dimension is not a multiple of 4

fabiencastan:


yurymalkov:
Thank you @fabiencastan! I'll check the PR during the weekend!

yurymalkov:
Thanks! I've already forgot that dims with d%4!=0 where not supported at all!

PR #245, opened 2020-08-29
Fix some compiler warnings

jlmelville:
This PR removes an initialization-order and several integer-kind comparison warnings.

I used the python bindings and `python setup.py test` to determine that I hadn't broken anything. I also didn't detect any slow down from using `size_t` instead of `int` to remove the integer comparison warnings. At the very least, the initialization reordering (line 29) should be totally safe.


jlmelville:
Closing this as it is pointlessly out-of-date.

PR #258, opened 2020-11-23
Use dict for Index serialization

dbespalov:
@yurymalkov @dyashuni this PR is still a WIP and untested. I should be a able to finish it tomorrow

### DIFF
--- a/python_bindings/bindings.cpp
+++ b/python_bindings/bindings.cpp
@@ -9,6 +9,7 @@
 #include <assert.h>
 
 namespace py = pybind11;
+using namespace pybind11::literals; // needed to bring in _a literal 
 
 /*
  * replacement for the openmp '#pragma omp parallel for' directive
@@ -72,6 +73,12 @@ inline void ParallelFor(size_t start, size_t end, size_t numThreads, Function fn
 
 
 }
+
+    inline void assert_true(bool expr, const std::string & msg) {
+      if (expr == false)
+        throw std::runtime_error("assert failed: "+msg);
+      return;
+    }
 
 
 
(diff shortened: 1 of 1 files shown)


dbespalov:
@yurymalkov PR is ready for the review. 

One item to potentially look into:  pybind11 complained when I tried returning dict as object's state, so I wrapped the parameter dict into tuple (of length one). I did not give it too much thought, tbh, so it might still be possible to avoid this.

python_bindings/bindings.cpp
```diff
@@ -345,140 +348,153 @@ class Index {
...
+      appr_alg->maxlevel_ = d["max_level"].cast<int>();
+      appr_alg->enterpoint_node_ = d["enterpoint_node"].cast<hnswlib::tableint>();
 
-      appr_alg->maxlevel_ = t[6].cast<int>();
-      appr_alg->enterpoint_node_ = t[7].cast<hnswlib::tableint>();
+      assert_true(appr_alg->maxM_ == d["max_M"].cast<size_t>(), "Invalid value of maxM_ ");
+      assert_true(appr_alg->maxM0_ == d["max_M0"].cast<size_t>(), "Invalid value of maxM0_ ");
+      assert_true(appr_alg->M_ == d["M"].cast<size_t>(), "Invalid value of M_ ");
+      assert_true(appr_alg->mult_ == d["mult"].cast<double>(), "Invalid value of mult_ ");
+      assert_true(appr_alg->ef_construction_ == d["ef_construction"].cast<size_t>(), "Invalid value of ef_construction_ ");
```
yurymalkov, line 486: I think the messages might confuse the user. I would add something like "Unpickle error:...".

yurymalkov (approved):
@dbespalov Thanks for an awesome contribution! 
One cosmetic change I would suggest is to extend the assert messages, so in case of a failure a user would better understand what has happened.

python_bindings/bindings.cpp
dbespalov: changed the error message to "Unpickle error"

dbespalov:
@yurymalkov good point, changed the error message. Thank you!

yurymalkov:
Great! Thank you so much!

PR #260, opened 2020-11-28
Move setup.py into root folder

dyashuni:
There is a build issue in Windows: symlink doesn't work in git. Therefore it is not possible to build python bindings on Windows. To fix it we can move the setup.py into the root folder.


yurymalkov:
Thanks for the PR! Can you please add symbolic links to the old locations for the compatibility with older scripts?
Can you also modify travis to do CI for windows (e.g. following https://docs.travis-ci.com/user/languages/python/)?

yurymalkov:
Please add the symbolic link to `setup.py`. It seems a link for `hnswlib` is not needed - just altering a `setup.py` file. Something like this:
```python
if "python_bindings" in os.path.basename(os.getcwd()): # compatibility when run in python_bindings
    source_files = ['./bindings.cpp']
    include_dirs = ['../hnswlib/']
else:
    source_files = ['./python_bindings/bindings.cpp']
    include_dirs = ['./hnswlib/']
```

dyashuni:
Done, please check.

yurymalkov:
Looks good to me! Thank you!

PR #262, opened 2020-12-06
Add pyproject.toml

groodt:
Modernizes the packaging slightly and follows the pybind11 guidelines:
https://pybind11.readthedocs.io/en/stable/compiling.html#pep-518-requirements-pip-10-required

Avoids the issues created by not specifying pybind11 in the `setup.py` which causes issues since it is imported:
https://github.com/groodt/hnswlib/blob/11a12199ed7d8c6ddb8910787809da138c94ada5/python_bindings/setup.py#L88


groodt:
Hmmm... doesn't actually work for local installs.

Issue #269, opened 2020-12-27
Python install error


davidmezzetti:
First off, thank you for this library, it works great and it's definitely fast.

I've run into the issue below for a while now when including hnswlib as a install_requires dependency in an empty virtual environment. Currently, the hnswlib install is falling back to a `python setup.py install` build but per the deprecation message below that will break with pip 21.0. 

_DEPRECATION: hnswlib was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. pip 21.0 will remove support for this functionality. A possible replacement is to fix the wheel build issue reported above. You can find discussion regarding this at https://github.com/pypa/pip/issues/8368._

This is the stacktrace when hnswlib is a dependency via install_requires. 

  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/pip-install-8sme64a_/hnswlib_db0256fc916d446cb5402165d7531712/setup.py", line 116, in <module>
      zip_safe=False,
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/setuptools/__init__.py", line 143, in setup
      return distutils.core.setup(**attrs)
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/distutils/core.py", line 148, in setup
      dist.run_commands()
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/distutils/dist.py", line 955, in run_commands
      self.run_command(cmd)
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/distutils/dist.py", line 974, in run_command
      cmd_obj.run()
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/wheel/bdist_wheel.py", line 299, in run
      self.run_command('build')
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/distutils/cmd.py", line 313, in run_command
      self.distribution.run_command(command)
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/distutils/dist.py", line 974, in run_command
      cmd_obj.run()
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/distutils/command/build.py", line 135, in run
      self.run_command(cmd_name)
    File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/distutils/cmd.py", line 313, in run_command
[...]

Issue #177, opened 2019-12-06
No pyproject.toml


andyalmonte:
`setup.py` imports `pybind11` and `numpy` and assumes that this will be available

https://github.com/nmslib/hnswlib/blob/0dcfb9111b3a9a349a86ec23f84b70705599f6ea/python_bindings/setup.py#L88-L89

There should be a pyproject.toml that requires both those modules to be installed before the setup.py is run

yurymalkov:
Hi @andyalmonte,

Can you make a pull request for that?

groodt:
This should close this issue:
https://github.com/nmslib/hnswlib/pull/274

PR #274, opened 2021-01-10
PEP-517 and PEP-518 support (add pyproject.toml)

groodt:
Closes: https://github.com/nmslib/hnswlib/issues/269 https://github.com/nmslib/hnswlib/issues/177

hnswlib does not use the recommended packaging approach for pybind11.
https://pybind11.readthedocs.io/en/stable/compiling.html#setup-helpers-pep518

The pyproject.toml file can specify the requirements necessary for the build backend (setuptools in this case) that are installed *before* the actual build takes place.

This should also make it easier to start packaging wheels etc if this project moves to a more modern packaging approach.


.gitignore
```diff
@@ -1,7 +1,8 @@
-hnswlib.egg-info/
```
groodt, line 1: I think the file was originally created on Windows? It had `\r` in the file. My only intended change was to add the `var/` ignore. I can remove this change if desired.

.travis.yml
```diff
@@ -30,8 +30,7 @@ jobs:
 
 install:
   - |
-    pip install -r requirements.txt
-    python setup.py install
+    python -m pip install .
```
groodt, line 33: `pip` will stop calling `python setup.py install` in future (after pip 21.x), it's better to install using `pip install .`

setup.py
```diff
@@ -3,17 +3,25 @@
 from setuptools.command.build_ext import build_ext
 import sys
 import setuptools
+import pybind11
```
groodt, line 6: These can now be safely imported due to the `pyproject.toml`

.gitignore
dyashuni: @groodt Yes, it was created on Windows.

setup.py
```diff
@@ -3,17 +3,25 @@
...
 
 __version__ = '0.4.0'
 
+
+include_dirs = [
+    pybind11.get_include(),
+    np.get_include(),
+]
+
 # compatibility when run in python_bindings
```
groodt, line 17: What is this compatibility for? I think removing it should be considered. Then it would be possible to start using [cibuildwheel](https://github.com/joerick/cibuildwheel) to build and distribute wheels for linux, osx, windows.

dyashuni: @groodt as @yurymalkov said there are some external scripts (outside the _nmslib/hnswlib_ repository, for example [Dockerfile.hnswlib](https://github.com/erikbern/ann-benchmarks/blob/master/install/Dockerfile.hnswlib)) that run the _setup.py_ from the _python_bindings_ folder. Therefore we added a check if _python_bindings_ in the current directory or not.

groodt: @dyashuni @yurymalkov I'll happily update that Dockerfile. It seems to be checking out a particular branch, so we could merge or rebase master into that branch after develop is merged? It might be worth thinking about pinning to a particular version / tag / sha as well.

If this PEP-517+518 change that I've proposed is merged, then it's even easier to use hnswlib as a dependency because minimal dependencies are needed on the host, `pip install hnswlib==x.y.z` will take care of installing the build time dependencies for building the sdist.

I thik it would be even better for `hnswlib` to provide wheels in pypi because then it's even easier for to use hnswlib as a dependency because no CPP toolchains etc are necessary. This of course isn't what I'm proposing in this PR. This PR is just making this project correctly specify it's build-time dependencies so that sdist can be built reliably without manual steps for users of hnswlib.

yurymalkov (commented):
Cool! Thanks! 
I wonder, will the current script (https://github.com/nmslib/hnswlib/blob/develop/Makefile) work for publishing pip packages?

groodt:
> I wonder, will the current script (https://github.com/nmslib/hnswlib/blob/develop/Makefile) work for publishing pip packages?

Yes, in this PR, the project will still publish an "sdist" (source distribution), not a "wheel" (binary distribution). However, now the "sdist" will be built correctly by consumers of this package without having to manually install `pip install pybind11 numpy setuptools` into their environment beforehand when the legacy fallback behavior of `pip` is removed.

Compare what happens with `pip install .` of the `develop` branch and `pip install hnswlib==0.40`:
```
Using legacy setup.py install for hnswlib, since package 'wheel' is not installed.
```

This message no longer appears if you try the install into a clean virtualenv of this branch.

You will see the `PEP-517` behavior kick in:
```
**Building wheels for collected packages: hnswlib
  Building wheel for hnswlib (PEP 517) ... done**
```

yurymalkov:
Got it! Thanks! I'll check it and merge. @dyashuni can you please also check?

dyashuni:
Sure, I'll check it.

dyashuni:
Currently in travis we build hnswlib twice. When we run `python -m pip install` at an installation phase and `python setup.py test` at a test phase. I wonder why `python setup.py test` rebuilds the lib it is installed.
Link to travis logs: https://travis-ci.org/github/nmslib/hnswlib/jobs/753778781

groodt:
> Currently in the travis we build hnswlib twice. When we run `python -m pip install` at an installation phase and `python setup.py test` at a test phase. I wonder why `python setup.py test` rebuilds the lib if it is installed.
> Link to travis logs: https://travis-ci.org/github/nmslib/hnswlib/jobs/753778781

Yes. I didn't really want to mention it, but `setup.py test` is also deprecated: https://github.com/pypa/setuptools/pull/1878

I've made the changes in the latest commit. See: https://github.com/nmslib/hnswlib/pull/274/commits/467c98f03d49f5995c96a2a513bb6cf88e719f07

Essentially, this project can do with a bit of refactoring in the way that it is packaged and the Python coding conventions to make it follow best-practices a bit.

I can raise a few more PRs to propose some changes in future if I find some time and if the project is interested.

setup.py
```diff
@@ -117,8 +114,7 @@ def build_extensions(self):
     url='https://github.com/yurymalkov/hnsw',
     long_description="""hnsw""",
     ext_modules=ext_modules,
-    install_requires=['pybind11>=2.0', 'numpy'],
```
groodt, line 120: See: https://github.com/pybind/python_example

setup.py
```diff
@@ -117,8 +114,7 @@ def build_extensions(self):
...
+    install_requires=['numpy'],
     cmdclass={'build_ext': BuildExt},
-    test_suite="python_bindings.tests",
```
groodt, line 122: Deprecated. See: pypa/setuptools#1878

setup.py
```diff
@@ -1,19 +1,28 @@
 import os
-from setuptools import setup, Extension
-from setuptools.command.build_ext import build_ext
 import sys
+
+import numpy as np
```
groodt, line 4: Dependencies sorted in all Python code according to PEP8 https://www.python.org/dev/peps/pep-0008/#imports

(I ran isort)

.travis.yml
```diff
@@ -30,9 +30,8 @@ jobs:
...
 
 script:
   - |
-    python setup.py test
```
groodt, line 38: Deprecated. See: pypa/setuptools#1878

python_bindings/tests/bindings_test.py
```diff
@@ -1,11 +1,13 @@
 import os
 import unittest
 
+import numpy as np
```
groodt, line 4: Dependencies sorted in all Python code according to PEP8 https://www.python.org/dev/peps/pep-0008/#imports

(I ran isort)

python_bindings/tests/bindings_test_labels.py
```diff
@@ -1,131 +1,127 @@
 import os
 import unittest
 
+import numpy as np
 
-class RandomSelfTestCase(unittest.TestCase):
-    def testRandomSelf(self):
-      for idx in range(16):
```
groodt, line 7: Fixed inconsistent indentation. It was a mix of 2 and 4 spaces in this file. Now it is 4 spaces.

yurymalkov (approved):
Looks good to me!

dyashuni:
Good changes, thank you very much @groodt!

dyashuni (approved):

dyashuni (approved):

dyashuni:
One thing to mention, now in Makefile `python3 setup.py sdist` doesn't work if _pybind11_ is not installed.

What is the best way to handle it @groodt ? Add an additional target in Makefile that will allow a user to install all required packages to build sdist, for example, `pip3 install pybind11 numpy setuptools` ?

groodt:
> One thing to mention, now in Makefile `python3 setup.py sdist` doesn't work if _pybind11_ is not installed.

Ah, you are correct. My apologies. I should have checked. Yes, these new PEP standards do encourage more publishing of wheels than sdist. We don't need to make this change now (or ever).

At least this inconvenience is only for the package uploaded to PyPI, which is presumably a limited number of people.

There is an official PEP517 builder that supports both sdist and wheels, see: https://github.com/pypa/build

I'll make this change to the Makefile shortly.

README.md
```diff
@@ -213,8 +213,9 @@ print("Recall for two batches:", np.mean(labels.reshape(-1) == np.arange(len(dat
 You can install from sources:
 ```bash
 apt-get install -y python-setuptools python-pip
-pip3 install pybind11 numpy setuptools
-python3 setup.py install
+git clone git@github.com:nmslib/hnswlib.git
```
dyashuni, line 216: Looks like that it should be changed to the `git clone https://github.com/nmslib/hnswlib.git`

dyashuni:
> python3 setup.py sdist

It fixes the issue, thank you @groodt !

README.md
groodt: Agree. The `https://` scheme is probably more user friendly for github users who haven't setup ssh keys and those that have will already know how to clone the source. 👍

dyashuni (approved):
@groodt Thank you!

yurymalkov (approved):
Looks great! Do you plan add something to this PR?

groodt:
Nope. I would be happy for it to be merged.

yurymalkov:
Cool! Thank you so much!
Also thanks to @dyashuni for reviewing!

groodt:
Thanks for the reviews @yurymalkov @dyashuni 

Is there likely to be a new release to PyPI anytime soon or are there more features waiting to land on `develop`?

yurymalkov:
Yes, we plan to release soon, there are significant changes.
Most likely, the following week.

groodt:
@yurymalkov No rush, just checking in, but any news on the next release date?

yurymalkov:
@groodt 
Done! Thanks!

groodt:
> @groodt
> Done! Thanks!

Awesome! Thanks again!

PR #289, opened 2021-02-25
Improved description of `add_items`

TakaakiFuruse:
This is just a suggestion of doc improvement.

I have made description of `labels` and `data_labels` together since...

1. There's no argument called `labels` for `add_items` func.
2. It felt like `labels` were a typo of `data_labels` from a previous [commit](https://github.com/nmslib/hnswlib/commit/5c20009894603810f1a3c3135e4bdf5d4025a782#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R37).

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ num_elements = 10000
 
 # Generating sample data
 data = np.float32(np.random.random((num_elements, dim)))
-data_labels = np.arange(num_elements)
+ids = np.arange(num_elements)
 
 # Declaring index
 p = hnswlib.Index(space = 'l2', dim = dim) # possible options are l2, cosine or ip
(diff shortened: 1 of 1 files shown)


yurymalkov:
Hi @TakaakiFuruse
Thank you for the PR!
 It seems both names are wrong, and the correct name should be `ids`...(https://github.com/nmslib/hnswlib/blob/master/python_bindings/bindings.cpp#L664). It seems there are also some other names that are wrong.
I would appreciate if you fix it and update the PR.

TakaakiFuruse:
@yurymalkov 

I am happy to fix my PR.

>  It seems there are also some other names that are wrong.

So for example,  

`add_items(data, data_labels, num_threads = -1)`

should be 

`add_items(data, ids, num_threads = -1)`

??

yurymalkov:
Yes, that would be nice!
Other stuff is also wrong. E.g. `mark_deleted` should have `label` instead of `data_label` .
`init_index` has wrong parameter order (`M` should come before `ef_construction` ).

TakaakiFuruse:
@yurymalkov 
FIXED!!

yurymalkov:
@TakaakiFuruse thank you so much!

PR #297, opened 2021-03-23
Throw an exception if passed an unrecognized space_name.

psobot:
Currently, `hnswlib` will segfault if called from Python with an unrecognized `space_name`, as `l2space` is left undefined. This PR proposes a fix for that by throwing an exception instead.


yurymalkov:
Great! Thanks for the fix!

PR #332, opened 2021-07-25
Add bindings to brute force index

alonre24:
This PR goal is allowing users that use the python bindings to have a nicer way of testing expected recall. This is done by adding the ability to create and use the "brute-force" index via python code, and compare HNSW search results to the actual results on any query.

The PR contains a description of the new API, along with a detailed example.


yurymalkov:
Great! Thank you so much for the work! (and sorry it took so long to check)

yurymalkov:
It failed windows build - fixed it in #346

PR #339, opened 2021-09-09
Added AVX512 support for space_l2 and space_ip.

slice4e:
Add AVX512 support to L2 and IP algorithms. 

Testing with Ann-Benchmarks shows measurable improvement on c5.4xlarge and m6i.4xlarge instances. 

c5.4xlarge results
fashion-mnist-784-euclidean
fashion-mnist-784-euclidean_10_euclidean
mnist-784-euclidean
mnist-784-euclidean_10_euclidean
nytimes-256-angular
nytimes-256-angular_10_angular
sift-128-euclidean
sift-128-euclidean_10_euclidean

m6i.4xlarge results: 
fashion-mnist-784-euclidean
fashion-mnist-784-euclidean_10_euclidean
mnist-784-euclidean
mnist-784-euclidean_10_euclidean
nytimes-256-angular
nytimes-256-angular_10_angular
sift-128-euclidean
sift-128-euclidean_10_euclidean

### DIFF
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -25,8 +25,10 @@
 
 #if defined(__GNUC__)
 #define PORTABLE_ALIGN32 __attribute__((aligned(32)))
+#define PORTABLE_ALIGN64 __attribute__((aligned(64)))
 #else
 #define PORTABLE_ALIGN32 __declspec(align(32))
+#define PORTABLE_ALIGN64 __declspec(align(64))
 #endif
 #endif
 
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -4,6 +4,8 @@
 #define USE_SSE
 #ifdef __AVX__
 #define USE_AVX
+#ifdef __AVX512__
+#define USE_AVX512
 #endif
 #endif
 #endif
@@ -14,6 +16,10 @@
 #include <stdexcept>
 #else
 #include <x86intrin.h>
+#endif
+
+#if defined(USE_AVX512)
+#include <immintrin.h>
 #endif
 
 #if defined(__GNUC__)
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -9,6 +9,7 @@
 #endif
 #endif
 #endif
+#endif
 
 #if defined(USE_AVX) || defined(USE_SSE)
 #ifdef _MSC_VER
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -4,7 +4,7 @@
 #define USE_SSE
 #ifdef __AVX__
 #define USE_AVX
-#ifdef __AVX512__
+#ifdef __AVX512F__
 #define USE_AVX512
 #endif
 #endif
--- a/hnswlib/space_l2.h
+++ b/hnswlib/space_l2.h
@@ -23,7 +23,10 @@ namespace hnswlib {
 
     // Favor using AVX512 if available.
     static float
-    L2SqrSIMD16Ext(void *pVect1, void *pVect2, size_t qty) {
+    L2SqrSIMD16Ext(const void *pVect1v, const void *pVect2v, const void *qty_ptr) {
+        float *pVect1 = (float *) pVect1v;
+        float *pVect2 = (float *) pVect2v;
+        size_t qty = *((size_t *) qty_ptr);
         float PORTABLE_ALIGN64 TmpRes[16];
         size_t qty16 = qty >> 4;
 
--- a/hnswlib/space_l2.h
+++ b/hnswlib/space_l2.h
@@ -24,7 +24,7 @@ namespace hnswlib {
     // Favor using AVX512 if available.
     static float
     L2SqrSIMD16Ext(float *pVect1, float *pVect2, size_t qty) {
-        float PORTABLE_ALIGN32 TmpRes[16];
+        float PORTABLE_ALIGN64 TmpRes[16];
         size_t qty16 = qty >> 4;
 
         const float *pEnd1 = pVect1 + (qty16 << 4);
--- a/hnswlib/space_l2.h
+++ b/hnswlib/space_l2.h
@@ -19,7 +19,38 @@ namespace hnswlib {
         return (res);
     }
 
-#if defined(USE_AVX)
+#if defined(USE_AVX512)
+
+    // Favor using AVX512 if available.
+    static float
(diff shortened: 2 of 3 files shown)


hnswlib/hnswlib.h
```diff
@@ -4,6 +4,8 @@
 #define USE_SSE
 #ifdef __AVX__
 #define USE_AVX
+#ifdef __AVX512__
```
yurymalkov, line 7: Hm. It looks like this `#ifdef` does not have a corresponding `#endif` and this seems to trigger the CI build fail for some systems.

hnswlib/space_l2.h
```diff
@@ -19,7 +19,38 @@ namespace hnswlib {
...
+        __m512 diff, v1, v2;
+        __m512 sum = _mm512_set1_ps(0);
+
+        while (pVect1 < pEnd1) {
+            v1 = _mm512_loadu_ps(pVect1);
+            pVect1 += 16;
+            v2 = _mm512_loadu_ps(pVect2);
+            pVect2 += 16;
+            diff = _mm512_sub_ps(v1, v2);
+            // sum = _mm512_fmadd_ps(diff, diff, sum);
```
yurymalkov, line 41: I wonder why the line is commented? Does it perform slower?

yurymalkov (changes requested):
Thanks you so much for the PR! 
It seems like there is #endif missing. Can you please check?
Also, I wonder if you've tested it with sustained multithreaded load? I heard AVX512 can cause cloud instances to overheat and throttle, so they can even perform worse.

hnswlib/space_l2.h
slice4e: Yes - I tested a fused multiply-add and it was slower for the AVX2 code. Was the same for the AVX512 code, so I decided to not use it.

hnswlib/hnswlib.h
slice4e: Yes, I did test with multiple concurrent processes running. We plugged the functions:  InnerProductSIMD16Ext and  L2SqrSIMD16Ext into a different application. The new application was implementing a brute-force vector similarity search.
The test was performed on c5.4xlarge on AWS. 

We can see that AVX512 is consistently better (lower execution time) than AVX2. We can also see that after about 8 parralel instances, the latency shoots up dramatically. We believe that this is due to memory latency. The brute-force algorithm is a memory scan and thus very memory bandwith intensive. After about 8 concurrent instances, I expect that we are saturating the memory bandwidth of the CPU. 

IP_brute_force
L2_brute_force
 
It is true that frequency goes down with AVX512 as compared with AVX2. Same is also true for AVX2 vs AVX. The number of cores executing AVX/AVX2/AVX512 code, determines the frequency. The 3 figures below, show the frequency tables depending on how many cores are active. As an example, I have circled a single point - assuming all cores active on a 8280 CPU (which is CascadeLake - same generation as used in c5 AWS instances). 

We can see that with all cores active, AVX code will run at 3.3GHz, AVX2 code will run at 2.9GHz, and AVX512 code will run at 2.4GHz. Even though, there is a frequency drop, we are operating on 2X the data at a time. So, if we spend enough time in the computation, then AVX512 will be a net benefit. In addition, AVX512 frequency is improving in subsequent generations - e.g. The m6i instance (IceLake ) is bridging the gap between AVX2 and AVX512 frequency. 

AVX_frequency
AVX2_frequency
AVX512_frequency

slice4e: > Hm. It looks like this `#ifdef` does not have a corresponding `#endif` and this seems to trigger the CI build fail for some systems.

Thank you for catching this. I have added the missing `#endif`

slice4e: The initial ann-bechmark results that I posted were with parallelism=3 on an c5.4xlarge instance. 
I just tested with parallelism=6 and the results are similar. I tried with parallelism=8, but we run out of memory on c5.4xlarge. 

image
image
image
image

yurymalkov: Thanks for the test and explanation!

yurymalkov:
I've tried to run the code and noticed that `__AVX512__` is not defined in gcc with `-march=native` (there are `__AVX512BW__`, `__AVX512F__`, `__AVX512VNNI__` and many others, probably corresponding to different subsets). 
I wonder, how do you test the code? Can you please add instructions on how to compile with AVX512 support?

slice4e:
> I've tried to run the code and noticed that `__AVX512__` is not defined in gcc with `-march=native` (there are `__AVX512BW__`, `__AVX512F__`, `__AVX512VNNI__` and many others, probably corresponding to different subsets).
> I wonder, how do you test the code? Can you please add instructions on how to compile with AVX512 support?

Hi - Sorry, this was a mistake on my part. For some reason, I did not realize that we are checking for the presence of AVX/AVX2/etc at compile time, and I assumed that the application using hnswlib is setting the defines. In highsight, that does not make sense...

I fixed it now. We are now checking for __AVX512F__ (which is the foundational set of AVX512 instructions.). All the instructions that we are using are part of __AVX512F__
For my performance testing, I had simply hardcoded the #define USE_AVX512 ... in the AVX512 tests. 


In addition, I fixed a bug which was causing a crash on some system due to not aligning the memory correctly for the store instruction (_mm512_store_ps). This instruction requires memory to be aligned to 64 bytes.

hnswlib/space_l2.h
```diff
@@ -19,7 +19,38 @@ namespace hnswlib {
         return (res);
     }
 
-#if defined(USE_AVX)
+#if defined(USE_AVX512)
+
+    // Favor using AVX512 if available.
+    static float
+    L2SqrSIMD16Ext(float *pVect1, float *pVect2, size_t qty) {
```
yurymalkov, line 26: Sorry for a long response. It seems `float` type here obstructs the compilation.
It should be `void *`

slice4e: > Sorry for a long response. It seems `float` type here obstructs the compilation. It should be `void *`

Thanks for catching this. I have fixed it.

yurymalkov: I've tried and there other errors with `constant  void *` to `void *` conversion (I guess it should be made const void * down the stack).
Can you please fix it and test it with `pip install .` and `python -m unittest discover --start-directory python_bindings/tests --pattern "*_test*.py"` (to run the tests) on a linux system.

slice4e: Thanks for the help! I fixed those errors. I was able to compile it and test it without errors following your instructions.

yurymalkov:
Thanks for the PR!!

PR #340, opened 2021-09-12
Expose CMake target for use via FetchContent

LTLA:
This allows other CMake projects to include HNSW via CMake's `FetchContent` approach, e.g.,

```
FetchContent_Declare(
  hnswlib
  GIT_REPOSITORY https://github.com/nmslib/hnswlib
  GIT_TAG master
)

FetchContent_MakeAvailable(hnswlib)

target_link_libraries(my_lib PUBLIC hnswlib)
```

A real-world example is available at https://github.com/LTLA/knncolle/blob/master/extern/CMakeLists.txt.

I moved the test building directives into a conditional that only triggers when actually building Hnswlib. This ensures that downstream projects are not saddled with the need to build the tests when all they want are the headers.

I also took the liberty of making `getDataByLabel` `const`, to indicate that it could be safely used in parallel frameworks that do not expect modifications to the index upon query; and I added a missing header for `<deque>`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,28 +1,27 @@
-cmake_minimum_required (VERSION 2.6)
-project (hnsw_lib)
+cmake_minimum_required (VERSION 3.14)
+project(hnsw_lib
+    LANGUAGES CXX)
 
-include_directories("${PROJECT_BINARY_DIR}")
+add_library(hnswlib INTERFACE)
+target_include_directories(hnswlib INTERFACE .) 
 
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
+    set(CMAKE_CXX_STANDARD 11)
 
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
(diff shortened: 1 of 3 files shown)


yurymalkov:
@LTLA thank you for the PR!
I wonder is minimal 3.14 version of cmake is a real necessity? E.g. I've check my mac and it is 3.10 and a google cloud instance has 3.13.4. Even colab has only 3.12.0. 
I guess this can cause compatibility issues.

LTLA:
That's... a good question. Not sure why I bumped it up to 3.14, maybe just a force of habit. I lowered it back down to 2.6 with no issues for **hnswlib** itself. I don't think downstream projects will care because they'll set their own minimum requirements.

The only thing to note is that more recent versions of CMake will complain if the minimum version is too low. Probably it could be bumped to 3.* without many problems, but that's not entirely relevant to this PR, so I'll leave that to you.

yurymalkov:
@LTLA, thank you! I'll try the change shortly!

LTLA:
Gentle nudge.

yurymalkov:
@LTLA Thank you for the PR!

PR #350, opened 2021-11-26
LazyIndex

urigoren:
Added a `LazyIndex` class with the same API as `Index`.
Useful when having multiple indices with lazy initialization.

This also fixes the super annoying bug that crashes `hnswlib` (without an exception) if calling `add_items` or `get_max_elements` before calling `init_index`.

This PR in its first version probably needs some work, I'd love to hear your feedback, thanks :-)


yurymalkov:
Hi @urigoren ,
Thanks!

One thing I would change is the name of the file, since it directly collides with hnswlib name, which might cause ambiguity if imported within the folder. 
Another thing is - adding a short description in the file will be helpful for others to understand its purpose.

urigoren:
Hi, @yurymalkov - done.
Could you please review ?

Thanks!

yurymalkov:
Looks good to me. Thanks!

Issue #292, opened 2021-03-08
Add runtime check for AVX with SSE2 build


yurymalkov:
This issue is related to https://github.com/conda-forge/hnswlib-feedstock/pull/11 . @yihming nicely volunteered to help!

Currently the used SIMD instruction set is used based on compile-time flags with instance calculation being on of the most critical parts. This makes binary distribution crash on SSE2 or underperform on AVX.

We can add an option to compile the AVX instructions for a SSE2 build and call those if AVX is available (it is safe to assume SSE2 is available on all x64 machines).

For checking availability of AVX something like this https://stackoverflow.com/questions/6121792/how-to-check-if-a-cpu-supports-the-sse3-instruction-set . 
SSE2 distance functions are already implemented in https://github.com/nmslib/hnswlib/blob/master/hnswlib/space_l2.h and https://github.com/nmslib/hnswlib/blob/master/hnswlib/space_ip.h (though only a single version is compiled due to preprocessor checks). Selection of the distance function is also implemented (e.g. https://github.com/nmslib/hnswlib/blob/master/hnswlib/space_ip.h#L252), but it is done based only on the vector length. 

Testing: modified version (compiled with SSE2 arch) should have outperform SSE2-only baseline running search on the same 1-10M built index (random vectors with say 128 dimension).

yihming:
Hi @yurymalkov ,

I think you can close this issue as it's added to hnswlib v0.6.1+.


Sincerely,
Yiming

yurymalkov:
Thank you!

PR #357, opened 2021-12-22
AVX runtime check

tony-kuo:
Regarding issue https://github.com/nmslib/hnswlib/issues/292
AVX runtime check: I've made a wrapper function for the simd functions and runtime check which to use.  Not the cleanest implementation but I think of a simpler way.

In addition, for issue https://github.com/conda-forge/hnswlib-feedstock/pull/11
I've tested building on one machine and installing in another with different AVX capabilities with both conda and pip.  The '-march=native' in setup.py is a factor in illegal instruction errors.  Taking it out solves this issue.  So I propose a environment flag 'HNSWLIB_NO_NATIVE', which is checked for in setup.py.  This will allow conda build recipes to choose not to use '-march=native'.

### DIFF
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -84,7 +84,7 @@ bool AVXCapable() {
         uint64_t xcrFeatureMask = xgetbv(_XCR_XFEATURE_ENABLED_MASK);
         avxSupported = (xcrFeatureMask & 0x6) == 0x6;
     }
-    return avxSupported;
+    return HW_AVX && avxSupported;
 }
 
 bool AVX512Capable() {
@@ -96,8 +96,8 @@ bool AVX512Capable() {
     cpuid(cpuInfo, 0, 0);
     int nIds = cpuInfo[0];
 
-    bool HW_AVX512F = false;    //  AVX512 Foundation
-    if (nIds >= 0x00000007) {
+    bool HW_AVX512F = false;
+    if (nIds >= 0x00000007) { //  AVX512 Foundation
         cpuid(cpuInfo, 0x00000007, 0);
         HW_AVX512F = (cpuInfo[1] & ((int)1 << 16)) != 0;
     }
@@ -108,13 +108,12 @@ bool AVX512Capable() {
     bool osUsesXSAVE_XRSTORE = (cpuInfo[2] & (1 << 27)) != 0;
     bool cpuAVXSuport = (cpuInfo[2] & (1 << 28)) != 0;
 
-    bool avxSupported = false;
     bool avx512Supported = false;
     if (osUsesXSAVE_XRSTORE && cpuAVXSuport) {
         uint64_t xcrFeatureMask = xgetbv(_XCR_XFEATURE_ENABLED_MASK);
         avx512Supported = (xcrFeatureMask & 0xe6) == 0xe6;
     }
-    return avx512Supported;
+    return HW_AVX512F && avx512Supported;
 }
 
 namespace hnswlib {
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -37,6 +37,86 @@
 #include <iostream>
 #include <string.h>
 
+// Adapted from https://github.com/Mysticial/FeatureDetector
+#define _XCR_XFEATURE_ENABLED_MASK  0
+#ifdef _WIN32
+void cpuid(int32_t out[4], int32_t eax, int32_t ecx){
+    __cpuidex(out, eax, ecx);
+}
+__int64 xgetbv(unsigned int x){
+    return _xgetbv(x);
+}
+#else
+#include <cpuid.h>
+#include <stdint.h>
+void cpuid(int32_t cpuInfo[4], int32_t eax, int32_t ecx) {
+    __cpuid_count(eax, ecx, cpuInfo[0], cpuInfo[1], cpuInfo[2], cpuInfo[3]);
+}
+
+uint64_t xgetbv(unsigned int index) {
+    uint32_t eax, edx;
+    __asm__ __volatile__("xgetbv" : "=a"(eax), "=d"(edx) : "c"(index));
+    return ((uint64_t)edx << 32) | eax;
+}
+#endif
+
+bool AVXCapable() {
+    int cpuInfo[4];
+
+    // CPU support
+    cpuid(cpuInfo, 0, 0);
+    int nIds = cpuInfo[0];
+
+    bool HW_AVX = false;
+    if (nIds >= 0x00000001) {
+        cpuid(cpuInfo, 0x00000001, 0);
+        HW_AVX = (cpuInfo[2] & ((int)1 << 28)) != 0;
+    }
+
+    // OS support
+    cpuid(cpuInfo, 1, 0);
+
+    bool osUsesXSAVE_XRSTORE = (cpuInfo[2] & (1 << 27)) != 0;
+    bool cpuAVXSuport = (cpuInfo[2] & (1 << 28)) != 0;
+
+    bool avxSupported = false;
+    if (osUsesXSAVE_XRSTORE && cpuAVXSuport) {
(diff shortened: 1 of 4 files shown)


yurymalkov:
Thank you! Will review it within the week.

hnswlib/space_ip.h
```diff
@@ -245,6 +251,41 @@ namespace hnswlib {
...
+        simdfunc_ = InnerProductSIMD4ExtSSE;
+    #endif
+        return simdfunc_(pVect1v, pVect2v, qty_ptr);
+    }
+
+    static float
+    InnerProductSIMD16Ext(const void *pVect1v, const void *pVect2v, const void *qty_ptr) {
+        DISTFUNC<float> simdfunc_;
+    #if defined(USE_AVX512)
+        if (AVX512Capable())
```
yurymalkov, line 272: It seems like it does a test for the AVX every distance calculation, which I guess would damage the performance.

yurymalkov (changes requested):
Thank you for the pull request! 
I wonder if you've looked at the performance compared to the base version? It seems that the avx check is performed on every distance calculation. I would expect it to significantly hurt the performance for small vectors (e.g. d=4, though I have not checked that).

I think an alternative here is to move the avx check to the Space constructor (https://github.com/nmslib/hnswlib/blob/master/hnswlib/space_ip.h#L286), so it will be done only once.

hnswlib/space_ip.h
tony-kuo: I'll take another look to see if I can only perform the check only once in the constructor. The difficulty in doing that is that the *ExtResiduals functions use the *Ext functions and are outside the class.

yurymalkov: Maybe this can be solved by adding a template parameter with avx support inside so it (so compiler would optimize the unused branches away)?

tony-kuo: I got a VM for windows and fixed windows build now.

yihming:
Hi @yurymalkov . Tony is helping us work on the AVX runtime check for hnswlib. And we just wonder if you could allow running the CI test workflows after his recent modifications. Thanks!


Sincerely,
Yiming

yurymalkov:
Hi @tony-kuo. Sure! Thanks!
I guess there is an option to run the CI tests in the forks of the repository as well, thought this requires turning on github actions.

hnswlib/space_ip.h
yurymalkov: Thank you! I'll check it. Hopefully, this weekend.

setup.py
```diff
@@ -74,8 +74,13 @@ class BuildExt(build_ext):
     """A custom build extension for adding compiler-specific options."""
     c_opts = {
         'msvc': ['/EHsc', '/openmp', '/O2'],
-        'unix': ['-O3', '-march=native'],  # , '-w'
+        #'unix': ['-O3', '-march=native'],  # , '-w'
+        'unix': ['-O3'],  # , '-w'
     }
+    if not os.environ.get("HNSWLIB_NO_NATIVE"):
+        if "conda" not in sys.version.lower():
```
yurymalkov, line 81: A quick question here - why not anaconda? 
It seems like it cancels the use of AVX for all anaconda python builds.

tony-kuo: I think you're right.  Originally I was trying to catch when it was building the binary in the conda-forge systems, but with the flag now, that isn't necessary.  I'll remove the "conda" check and resubmit for review.

hnswlib/space_l2.h
```diff
@@ -208,7 +213,17 @@ namespace hnswlib {
...
-        #if defined(USE_SSE) || defined(USE_AVX) || defined(USE_AVX512)
+    #if defined(USE_SSE) || defined(USE_AVX) || defined(USE_AVX512)
+        #if defined(USE_AVX512)
+            if (AVX512Capable())
+                L2SqrSIMD16Ext = L2SqrSIMD16ExtAVX512;
+            else if (AVXCapable())
+                L2SqrSIMD16Ext = L2SqrSIMD16ExtAVX;
+        #elif defined(USE_AVX)
+            if (AVXCapable())
+                L2SqrSIMD16Ext = L2SqrSIMD16ExtAVX;
```
yurymalkov, line 224: This line overwrites the AVX512 selection above, so AVX512 is not used if it is supported. The same thing is for the inner product distance.

yurymalkov: That leads to AVX512 code never running

tony-kuo: The logic uses an "else if", so if AVX512 was compiled then it checks at run time to see if it is able to run, if not it will use AVX.  It will only go into the "USE_AVX" code block if it is not compiled on an AVX512 capable machine. This was confirmed with print statements showing it used the AVX512 code block.

yurymalkov: Ah. Ok. The bug is only in inner product then

hnswlib/space_ip.h
```diff
@@ -283,6 +292,19 @@ namespace hnswlib {
         InnerProductSpace(size_t dim) {
             fstdistfunc_ = InnerProduct;
     #if defined(USE_AVX) || defined(USE_SSE) || defined(USE_AVX512)
+        #if defined(USE_AVX512)
+            if (AVX512Capable())
+                InnerProductSIMD16Ext = InnerProductSIMD16ExtAVX512;
+            else if (AVXCapable())
+                InnerProductSIMD16Ext = InnerProductSIMD16ExtAVX;
+        #endif
+        #if defined(USE_AVX)
```
yurymalkov, line 301: I guess here should be `#elif`

tony-kuo: Ah yeah.  I tried to combine the check for both 16Ext and 4Ext.  But because 4Ext doesn't use AVX512 that led to a mistake.  If you do implement AVX512 in the future for 4Ext, then they can be combined in the runtime check.  I'll fix that and trigger another review.

yurymalkov (commented):
There is one more compatibility concern and question about a check for cpuid logic.
In other places looks good to me! 

I also (somewhat strangely) do not see much changes between AVX and AVX512 speed in my tests (and I've checked that different implementations are running) - I wonder what is your experience?

hnswlib/hnswlib.h
```diff
@@ -15,8 +15,25 @@
...
 #include <stdexcept>
+#include "cpu_x86.h"
+void cpu_x86::cpuid(int32_t out[4], int32_t eax, int32_t ecx) {
+    __cpuidex(out, eax, ecx);
+}
+__int64 xgetbv(unsigned int x) {
+    return _xgetbv(x);
+}
 #else
 #include <x86intrin.h>
```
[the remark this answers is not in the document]
tony-kuo: I also do not have non-x86 machine to test on unfortunately. The current avx check functions are encapsulated in the top level '#if defined(USE_AVX) || defined(USE_SSE)' and all the calls to the function are also in USE_AVX or USE_AVX512.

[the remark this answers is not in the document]
tony-kuo: I also do not see much speed difference between AVX and AVX512.  So I went back to the original code (the current master branch).  The performance with the runtime check vs the original code is the same.  Then I commented out the AVX512 code in space_l2.h and tested.  There was also no change that I saw with AVX vs AVX512 in the original code.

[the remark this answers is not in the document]
tony-kuo: I've also doubled check that the avx check functions work on linux and mac.  And actually, the avx check functions are already in the the top level "#if defined(USE_AVX) || defined(USE_SSE)"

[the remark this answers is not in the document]
yurymalkov: Yeah. You are right! Thanks for checking!

yurymalkov:
Thank you so much for the contribution!
The AVX512 speed seems weird, but that is outside of the scope of this PR. Will check the reason sepately.

PR #359, opened 2022-01-04
support for macOS armv8.5-a (MBP M1 Pro/Max)

jianshu93:
With python 3.9.9 for MBP armv8.5-a, this commit could allow successfully installation via "pip install ." in the directory.


yurymalkov:
It seems like this will break compatibility with other systems.

jianshu93:
Make it a seperate branch called arm？

yurymalkov:
This can be an option. Another option is to make a check in the script for armv.

yurymalkov:
closed as #363 is merged to develop. Please let me know if that is not ok.

PR #362, opened 2022-01-19
Throw exception instead of segfaulting when passing a scalar to get_items.

psobot:
Before this change:

```python
In [10]: i.get_items(1)
[1]    97271 segmentation fault  ipython
```

After this change:

```python
In [5]: i.get_items(1)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-6cfae0b544de> in <module>
----> 1 i.get_items(1)

ValueError: get_items accepts a list of indices and returns a list of vectors
```


yurymalkov:
Hi @psobot. Thanks for the PR!  It seems that one of the tests is failing. Can you please look at it?

yurymalkov:
@psobot Thank you!

PR #364, opened 2022-01-27
Add other data storage types to Python bindings.

psobot:
This is a bit of a big one - happy to break this into smaller PRs if that would be useful. This pull request:
 - Adds new Python bindings for indexes with different data storage types:
   - `DoubleIndex` (float64)
   - `Int8Index`
   - `UInt8Index`
   - `Int16Index`
   - `UInt16Index`
 - Adds another template argument to subclasses  of `SpaceInterface`: `data_t`, to specify the data storage type used.
 - Changes `get_items` to return a Numpy array instead of a `List[List[data_t]]`.
 - Adds templated `InnerProduct` and `L2Sqr` distance comparison functions that auto-unroll and auto-vectorize their inner loops ([see Godbolt](https://godbolt.org/z/baf1oo647)). This allows us to use different data types without manually having to write every comparison function. (This _might_ make the manual SIMD functions obsolete, although I've left them in there for now out of an abundance of caution.)
 - Extends test coverage to cover these new classes.

I'm not 100% sure if this is the best way to architect this API; in particular, should the data type be a property of `Index`, set at creation time, rather than a separate class of `Index`? (For now, it's the latter.)

The existing tests seem to pass, and the new data types allow for smaller index files on disk (in situations where reduced-precision is acceptable):
image

The new `int8` and `uint8` data types even seem to perform about 60% faster in the best case (when using 1024 dimensions):
image

### DIFF
--- a/python_bindings/bindings.cpp
+++ b/python_bindings/bindings.cpp
@@ -848,7 +848,7 @@ template<> const std::string typeName<unsigned char>() { return "uint8"; }
 template<> const std::string typeName<short>() { return "int16"; }
 template<> const std::string typeName<unsigned short>() { return "uint16"; }
 template<> const std::string typeName<int>() { return "int32"; }
-template<> const std::string typeName<uint>() { return "uint32"; }
+template<> const std::string typeName<unsigned int>() { return "uint32"; }
 template<> const std::string typeName<float>() { return "float32"; }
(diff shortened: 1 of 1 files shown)


patelprateek:
@psobot Curios if you are still working on this ?
With this change , do you see performance impact on the old fp32 distance computation ?

PR #382, opened 2022-06-07
Update removePoint in bruteforce.h

bongmo:
- When I add some points and delete all points, unordered_map remains item (bug)

### DIFF
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -78,7 +78,8 @@ namespace hnswlib {
             dict_external_to_internal.erase(cur_external);
 
             labeltype label=*((labeltype*)(data_ + size_per_element_ * (cur_element_count-1) + data_size_));
-            dict_external_to_internal[label]=cur_c;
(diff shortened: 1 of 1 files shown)


hnswlib/bruteforce.h
```diff
@@ -78,7 +78,8 @@ namespace hnswlib {
             dict_external_to_internal.erase(cur_external);
 
             labeltype label=*((labeltype*)(data_ + size_per_element_ * (cur_element_count-1) + data_size_));
-            dict_external_to_internal[label]=cur_c;
+            if (cur_external != label)
```
yurymalkov, line 81: I guess a check for `cur_element_count>1` would be easier to read.
Also I think copying of the element in the next lines do not make any sense so should be in the same `if`

yurymalkov:
@bongmo thanks for the PR! That does seem like a bug. 
A small suggestion would be to alter the change so it would be a bit easier to read (mentioned in the code comment).

PR #383, opened 2022-06-08
Fix global linkages

MasterAler:
Greetings!

First of all, thank you for your library, it's awesome and useful. 
But here's a tiny fix I dare to propose: as soon some of the headers contain definitions without `static` specified, could you add it to them?
Currenlty those definitions have global linkage, which means that in certain cases someone is likely to get "multiple definitions" error, I can provide you a MCVE of that here or in a separate issue if you like.

The fix is small and should not affect performance, tests pass fine.

@yurymalkov , what'd you say?

### DIFF
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -16,20 +16,20 @@
 #include <intrin.h>
 #include <stdexcept>
 #include "cpu_x86.h"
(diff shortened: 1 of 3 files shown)


yurymalkov:
Hi @MasterAler,

Thanks! Yeah, I do not think it might cause any problem. Will merge in a few days.

PR #396, opened 2022-07-30
initialize fields in constructor

### LINKED ISSUE #393: valgrind problems when searching a reloaded index
I recently attempted to update the R bindings for hnswlib but there is a valgrind problem of the form:

```
==217201== Conditional jump or move depends on uninitialised value(s)
==217201==    at 0x188D9AB5: hnswlib::HierarchicalNSW<float>::searchKnn(void const*, unsigned long) const (packages/tests-vg/RcppHNSW/src/../inst/include/hnswalg.h:1150)
```

This should also be triggerable via the second Python bindings example on the current README (the code after "An example with updates after serialization/deserialization").

The problem line: 
https://github.com/nmslib/hnswlib/blob/443d667478fddf1e13f2e06b1da4e1ec3a9fe716/hnswlib/hnswalg.h#L1150 

I believe the issue is that not all `HierarchicalNSW` constructors are initializing all of their data members. Specifically `num_deleted_` is a problem: while it can be incremented in `loadIndex` under circumstances where `isMarkDeleted` returns true, it's never actually assigned. In the example above, there isn't any deletion so the problem doesn't become apparent until `searchKnn` is invoked.

I haven't checked for the validity of other class member variables so other problems may be lurking. I can certainly help with providing a PR or helping with testing, but I haven't updated the R bindings for a couple of years so more recent functionality isn't tested (yet). So this will for sure need some extra eyes on it.

jlmelville:
Fixes #393.

Testing: I ran `make test` and got no different results between the output in this PR and the current state of `develop`. That said I do see these warnings in the output (these are *also* present in `develop`):

```
.Adding all elements (10000)
.Running pickle tests for <hnswlib.Index(space='cosine', dim=32)>
.Running pickle tests for <hnswlib.Index(space='ip', dim=16)>
Warning: 1 labels are missing from ann results (k=25, err_thresh=5)
Warning: 1 labels are missing from ann results (k=25, err_thresh=5)
Warning: 1 labels are missing from ann results (k=25, err_thresh=5)
Warning: 1 ann distance values are different from brute-force values (total # of values=2500, dists_thresh=50)
Warning: 1 labels are missing from ann results (k=25, err_thresh=5)
Warning: 1 labels are missing from ann results (k=25, err_thresh=5)
Warning: 1 labels are missing from ann results (k=25, err_thresh=5)
Warning: 1 ann distance values are different from brute-force values (total # of values=2500, dists_thresh=50)
```

I am not sure if this is an actual problem or not?

Also, I confirmed that the valgrind error disappears via testing through the R bindings I maintain.

### DIFF
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -8,10 +8,14 @@ namespace hnswlib {
     template<typename dist_t>
     class BruteforceSearch : public AlgorithmInterface<dist_t> {
     public:
-        BruteforceSearch(SpaceInterface <dist_t> *s) {
+        BruteforceSearch(SpaceInterface <dist_t> *s) : data_(nullptr), maxelements_(0), 
+        cur_element_count(0), size_per_element_(0), data_size_(0), 
+        dist_func_param_(nullptr) {
 
         }
-        BruteforceSearch(SpaceInterface<dist_t> *s, const std::string &location) {
+        BruteforceSearch(SpaceInterface<dist_t> *s, const std::string &location) : 
+        data_(nullptr), maxelements_(0), cur_element_count(0), size_per_element_(0), 
+        data_size_(0), dist_func_param_(nullptr) {
             loadIndex(location, s);
         }
 
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -17,15 +17,34 @@ namespace hnswlib {
     class HierarchicalNSW : public AlgorithmInterface<dist_t> {
     public:
         static const tableint max_update_element_locks = 65536;
-        HierarchicalNSW(SpaceInterface<dist_t> *s) {
+        HierarchicalNSW(SpaceInterface<dist_t> *s) : 
+        max_elements_(0), cur_element_count(0), size_data_per_element_(0), 
+        size_links_per_element_(0), num_deleted_(0), M_(0), maxM_(0), maxM0_(0), 
+        ef_construction_(0), mult_(0.0), revSize_(0.0), maxlevel_(0), 
+        visited_list_pool_(nullptr), enterpoint_node_(0), size_links_level0_(0), 
+        offsetData_(0), offsetLevel0_(0), 
+        data_level0_memory_(nullptr), 
+        linkLists_(nullptr), data_size_(0), label_offset_(0), 
+        dist_func_param_(nullptr), metric_distance_computations(0), metric_hops(0), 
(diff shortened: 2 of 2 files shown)


yurymalkov:
Cool! Thanks!

Before merging, I wonder if the constructor initialization is the preferred way (compared to C++ 11 initialization at the variable declaration)?

jlmelville:
> Before merging, I wonder if the constructor initialization is the preferred way (compared to C++ 11 initialization at the variable declaration)?

I can do that. The body of the third constructor `HierarchicalNSW(SpaceInterface<dist_t> *, size_t, size_t, size_t, size_t)`  (not changed in this PR) sets some of those fields to other values (e.g. `ef_construction` and `M_`). Would you want those fields also changed to default member initializers (presumably paying the minimal price of setting them twice) or leave those ones alone?

If we are talking constructor improvements then that third constructor would gain the most from changes, as it is the least C++ idiomatic as it sets most of its fields in the constructor body rather than using initializer lists. This PR will leave the constructor styles a bit inconsistent which might be confusing for readers. But I wanted to keep the scope of this PR as small as I could, especially as `clang-tidy` has a *lot* to say about converting hnswlib to C++11 form and it's hard to know where to stop once you start down that path...

Anyway let me know if you would like me to go ahead and use the default initialization approach (I see that `clang-tidy` may have missed some fields that should be default initialized).

yurymalkov:
Thanks! I am not a C++ expert, so not sure what is the best.
From my experience with other languages (java mostly) it is convenient to have declaration and default initialization at the same point of code (e.g. adding the default values in the block https://github.com/nmslib/hnswlib/blob/master/hnswlib/hnswalg.h#L88-L125 and not touching the existing constructors - they would only override the default values), so it would be also visible if some of the variables are not initialized without cross-checking. I thought that was introduced in C++11 (which is already used in hnswlib), though not sure if that is the right thing to do.

jlmelville:
@yurymalkov I have made the changes to use direct initialization for the non-class-type fields like `int`s and `size_t`s (I have also used braces for the initialization which seems to be preferred by the [C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines)). I believe more complex member types like `std::vector` will be default initialized without having to do anything but I am happy to be corrected on that.

Valgrind (via R) is still happy and `make test` gives the same output.

yurymalkov:
Great! Thank you so much!

PR #401, opened 2022-08-09
Add rust implementation

jianshu93:
A beautiful Rust implementation. I have run benchmarks for various dataset with very good performance and recall. More metric distance are supported including hamming, Hellinger distance, Jensen-Shannon distance between probability distributions (f32 and f64).  There is  A Trait to enable the user to implement its own distances. It takes as data slices of types T satisfying T:Serialize+Clone+Send+Sync.


yurymalkov:
Thanks!

Issue #366, opened 2022-02-01
creating ann index with membership bits 


prateekpatelsc:
@yurymalkov : I have a use case where i want to build ann index and also support filtering operation based on some tags .
The tags do not influence the distance or graph structure (during building index) , but during search operation we want to filter elements based on membership or filtering by tags associated. 
Two approaches i am thinking of : 
1) It seems like search operation can be similar as it is used to ignore deleted items. 
2) Add items with some added bit vector that indicates tags  
Build index with a custom distance function , where during indexing the usual distance metric is used (ignoring the bit vector), but while searching the distance function uses the added bit vector to decide eligibility , if eligibility == true return the original distance otherwise return some large distance (max float) . 

Do you observe any issues i.e any scenario in eligible items may get returned as topK ?
Any other approaches to support such filtering ?

TechnikEmpire:
I'm a little confused by this because the template parameters specifically enable you to associate a type with each entry so you have full control to implement what you're wanting here. If you're talking about modifying the vectors somehow I don't see how that's possible.

Take your unique index and build out a secondary structure where you do whatever custom logic you want there, no? Maybe I'm completely missing it here, just my $0.02.

prateekpatelsc:
I am not talking about modifying any vector . 
I want to build an index that also supports filtering i.e give me topK neighbors which have tags in [.. tag ids] 

There are 2 ways to implement this 
- similar to deleted items , if a topk neighbour is not satisfying a tag membership , it gets ingnored during graph traversal 
- second approach is add membership bit vector along with embedding , create index based on normal distance metric , but during search if membership is false return a large distance else return original distance , this way the topk neightbours will be most probably the correct filtered items

TechnikEmpire:
Yeah so what I'm saying is why you don't simply have a secondary element, where your index is std::uint64_t, something like:

std::unordered_map<std::uint64_t, MyFilterData>

And simply fetch KNN, use the returned indices to check your extra filter data and pop the elements that don't make the cut. This is precisely what you're asking for, you just want them to implement it and build it into the library. It is un-possible for this to be implemented any other way.

As a user of this library the last thing I want is to incur performance penalties for no reason, but that's my $0.02. Also just a disclaimer here I'm just a guy who starred the repo and uses this library to great effect.

prateekpatelsc:
may be i am not phrasing it properly.
[...]

PR #402, opened 2022-08-15
Filter elements with an optional filtering function

kishorenc:
Fixes https://github.com/nmslib/hnswlib/issues/366

**Summary of the change**

An optional filtering function can be specified as a template parameter that determines if a given ID should be picked. The default is a function that allows all labels via a`return true` -- this will be optimised away by the compiler, so there will be no extra cost for those who don't provide a filtering function. 

The use of a templated filter function also ensures that the filtering logic can be entirely inlined by the compiler, and is an implementation detail, instead of forcing a `std::set<>` of allowed IDs as discussed in #366.

I've added a test that asserts on both the brute force and hnsw implementations. Verified that existing search knn tests pass.

### DIFF
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -5,7 +5,7 @@
 #include <algorithm>
 
 namespace hnswlib {
-    template<typename dist_t, typename filter_func_t=FILTERFUNC>
+    template<typename dist_t, typename filter_func_t=FilterFunctor>
     class BruteforceSearch : public AlgorithmInterface<dist_t,filter_func_t> {
     public:
         BruteforceSearch(SpaceInterface <dist_t> *s) : data_(nullptr), maxelements_(0), 
@@ -92,7 +92,7 @@ namespace hnswlib {
 
 
         std::priority_queue<std::pair<dist_t, labeltype >>
-        searchKnn(const void *query_data, size_t k, filter_func_t isIdAllowed=allowAllIds) const {
+        searchKnn(const void *query_data, size_t k, filter_func_t& isIdAllowed=allowAllIds) const {
             std::priority_queue<std::pair<dist_t, labeltype >> topResults;
             if (cur_element_count == 0) return topResults;
             for (int i = 0; i < k; i++) {
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -102,7 +102,7 @@ namespace hnswlib {
                     topResults.push(std::pair<dist_t, labeltype>(dist, label));
                 }
             }
-            dist_t lastdist = topResults.top().first;
+            dist_t lastdist = topResults.empty() ? std::numeric_limits<dist_t>::max() : topResults.top().first;
             for (int i = k; i < cur_element_count; i++) {
                 dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
                 if (dist <= lastdist) {
@@ -112,7 +112,10 @@ namespace hnswlib {
                     }
                     if (topResults.size() > k)
                         topResults.pop();
-                    lastdist = topResults.top().first;
+
+                    if (!topResults.empty()) {
+                        lastdist = topResults.top().first;
+                    }
                 }
 
             }
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -5,8 +5,8 @@
 #include <algorithm>
 
 namespace hnswlib {
-    template<typename dist_t>
-    class BruteforceSearch : public AlgorithmInterface<dist_t> {
+    template<typename dist_t, typename filter_func_t=FILTERFUNC>
+    class BruteforceSearch : public AlgorithmInterface<dist_t,filter_func_t> {
     public:
         BruteforceSearch(SpaceInterface <dist_t> *s) : data_(nullptr), maxelements_(0), 
         cur_element_count(0), size_per_element_(0), data_size_(0), 
@@ -92,20 +92,24 @@ namespace hnswlib {
 
 
         std::priority_queue<std::pair<dist_t, labeltype >>
-        searchKnn(const void *query_data, size_t k) const {
+        searchKnn(const void *query_data, size_t k, filter_func_t isIdAllowed=allowAllIds) const {
             std::priority_queue<std::pair<dist_t, labeltype >> topResults;
             if (cur_element_count == 0) return topResults;
             for (int i = 0; i < k; i++) {
                 dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
-                topResults.push(std::pair<dist_t, labeltype>(dist, *((labeltype *) (data_ + size_per_element_ * i +
-                                                                                    data_size_))));
+                labeltype label = *((labeltype*) (data_ + size_per_element_ * i + data_size_));
+                if(isIdAllowed(label)) {
+                    topResults.push(std::pair<dist_t, labeltype>(dist, label));
+                }
             }
             dist_t lastdist = topResults.top().first;
             for (int i = k; i < cur_element_count; i++) {
                 dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
                 if (dist <= lastdist) {
-                    topResults.push(std::pair<dist_t, labeltype>(dist, *((labeltype *) (data_ + size_per_element_ * i +
-                                                                                        data_size_))));
+                    labeltype label = *((labeltype *) (data_ + size_per_element_ * i + data_size_));
+                    if(isIdAllowed(label)) {
+                        topResults.push(std::pair<dist_t, labeltype>(dist, label));
+                    }
                     if (topResults.size() > k)
                         topResults.pop();
                     lastdist = topResults.top().first;
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -93,12 +93,14 @@ namespace hnswlib {
 
         std::priority_queue<std::pair<dist_t, labeltype >>
         searchKnn(const void *query_data, size_t k, filter_func_t& isIdAllowed=allowAllIds) const {
+            assert(k <= cur_element_count);
             std::priority_queue<std::pair<dist_t, labeltype >> topResults;
             if (cur_element_count == 0) return topResults;
+            bool is_filter_disabled = std::is_same<filter_func_t, decltype(allowAllIds)>::value;
             for (int i = 0; i < k; i++) {
                 dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
                 labeltype label = *((labeltype*) (data_ + size_per_element_ * i + data_size_));
-                if(isIdAllowed(label)) {
+                if(is_filter_disabled || isIdAllowed(label)) {
                     topResults.push(std::pair<dist_t, labeltype>(dist, label));
                 }
             }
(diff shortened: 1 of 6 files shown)


hnswlib/bruteforce.h
```diff
@@ -92,20 +92,24 @@ namespace hnswlib {
...
             for (int i = 0; i < k; i++) {
                 dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
-                topResults.push(std::pair<dist_t, labeltype>(dist, *((labeltype *) (data_ + size_per_element_ * i +
-                                                                                    data_size_))));
+                labeltype label = *((labeltype*) (data_ + size_per_element_ * i + data_size_));
+                if(isIdAllowed(label)) {
+                    topResults.push(std::pair<dist_t, labeltype>(dist, label));
+                }
             }
             dist_t lastdist = topResults.top().first;
```
dyashuni, line 105: now topResults may be empty. Need to add a check here.
```
if (topResults.size() > 0)
{
    lastdist = topResults.top().first;
}
else
{
    lastdist = std::numeric_limits<dist_t>::max();
}
```

hnswlib/bruteforce.h
```diff
@@ -92,20 +92,24 @@ namespace hnswlib {
...
                 if (dist <= lastdist) {
-                    topResults.push(std::pair<dist_t, labeltype>(dist, *((labeltype *) (data_ + size_per_element_ * i +
-                                                                                        data_size_))));
+                    labeltype label = *((labeltype *) (data_ + size_per_element_ * i + data_size_));
+                    if(isIdAllowed(label)) {
+                        topResults.push(std::pair<dist_t, labeltype>(dist, label));
+                    }
                     if (topResults.size() > k)
                         topResults.pop();
                     lastdist = topResults.top().first;
```
dyashuni, line 115: Need to add a check here as well

```
if (topResults.size() > 0)
{
    lastdist = topResults.top().first;
}
```

dyashuni:
Thank you Kishore for implementing this feature!
My major concern is that isIdAllowed works with internal IDs that are hidden from users. Does it make sense to modify it to work with external IDs?

kishorenc:
@dyashuni 

Thank you for the review.

1. Good catch regarding the use of internal ID. I've now fixed the code to send the label to the filtering function and also updated the test to specifically assert for this issue by ensuring that the labels are not the same as the internal ID sequence.

2. I've also addressed the other issue about `topResults` empty checks. Added test for this one as well.

hnswlib/hnswalg.h
```diff
@@ -247,7 +247,7 @@ namespace hnswlib {
             std::priority_queue<std::pair<dist_t, tableint>, std::vector<std::pair<dist_t, tableint>>, CompareByFirst> candidate_set;
 
             dist_t lowerBound;
```
dyashuni, line 249: Can we add the following micro optimisation to not call isIdAllowed at all if filtering is disabled:

```
bool is_filter_disabled = isIdAllowed == allowAllIds;
if ((!has_deletions || !isMarkedDeleted(ep_id)) && (is_filter_disabled || isIdAllowed(getExternalLabel(ep_id)))) {
```

hnswlib/hnswalg.h
```diff
@@ -307,7 +307,7 @@ namespace hnswlib {
                                          _MM_HINT_T0);////////////////////////
 #endif
 
-                            if ((!has_deletions || !isMarkedDeleted(candidate_id)) && isIdAllowed(candidate_id))
+                            if ((!has_deletions || !isMarkedDeleted(candidate_id)) && isIdAllowed(getExternalLabel(candidate_id)))
```
dyashuni, line 310:
```
if ((!has_deletions || !isMarkedDeleted(ep_id)) && (is_filter_disabled || isIdAllowed(getExternalLabel(ep_id)))) {
```

kishorenc:
@dyashuni 

I just discovered a gotcha with the current filtering function interface: there's no way to pass some kind of application context. For e.g. if I have a set of IDs to check against, the static filter function will not be able to access that state. I'm going to think about how to pass a "context" to the filtering function.

hnswlib/bruteforce.h
```diff
@@ -92,20 +92,24 @@ namespace hnswlib {
 
 
         std::priority_queue<std::pair<dist_t, labeltype >>
-        searchKnn(const void *query_data, size_t k) const {
+        searchKnn(const void *query_data, size_t k, filter_func_t isIdAllowed=allowAllIds) const {
```
dyashuni, line 95: Would not it better to pass NULL as a default value to show that we do not do filtering by default (see https://github.com/nmslib/hnswlib/pull/402/commits/ad3440c83555d9a76eef0e23bc6505c86b026716#r954851933)?

`filter_func_t isIdAllowed=NULL`

dyashuni:
> @dyashuni
> 
> I just discovered a gotcha with the current filtering function interface: there's no way to pass some kind of application context. For e.g. if I have a set of IDs to check against, the static filter function will not be able to access that state. I'm going to think about how to pass a "context" to the filtering function.

Yes it is true. I thought that maybe we have some instruments in C++ to dynamically create functions with required parameters (e.g. variable filtering range) and pass pointers to such functions

Like in Python
https://stackoverflow.com/questions/803616/passing-functions-with-arguments-to-another-function-in-python
when we can pass functions with parameters using partial module or _lambda_ functions

dyashuni:
May be the best way is to pass a pointer to an instance of some class instead of pointer to a function.

kishorenc:
@dyashuni 

1. I've refactored the code to use a functor because function objects tend to get inlined very well by compilers (see [this](https://stackoverflow.com/a/40810103/131050)). 
2. Added an example in the test to show how a custom filtering function (with state) can be used.
3. I've also added the micro optimization as a compile time check

dyashuni:
@kishorenc thank you for the updates!

Could you please add the `searchKnnWithFilter_test` test to the CI?
We have just added cpp tests to the CI. You need to rebase and pick up the latest changes from the develop branch.
Then modify the `.github\workflows\build.yml` file, add `./searchKnnWithFilter_test` to the step `Test`


```
- name: Test
  run: |
    cd build
    ./searchKnnCloserFirst_test
    ./searchKnnWithFilter_test  <----- add this line
    ./test_updates
    ...
```

hnswlib/hnswalg.h
```diff
@@ -307,7 +308,8 @@ namespace hnswlib {
                                          _MM_HINT_T0);////////////////////////
 #endif
 
-                            if ((!has_deletions || !isMarkedDeleted(candidate_id)) && isIdAllowed(getExternalLabel(candidate_id)))
+                            is_filter_disabled = std::is_same<filter_func_t, decltype(allowAllIds)>::value;
```
dyashuni, line 311: I think we can remove this line as above we assigned value to `the is_filter_disabled` variable.

dyashuni:
@kishorenc Thank you!

In windows I get the following error:
searchKnnWithFilter_test.cpp(151,20): error C3615: constexpr function 'CustomFilterFunctor::operator ()' cannot result in a constant expression

Do we need to remove the `constexpr` keyword ?

kishorenc:
I've removed it now, please check again.

dyashuni:
Great, thank you!

hnswlib/bruteforce.h
```diff
@@ -92,23 +92,30 @@ namespace hnswlib {
 
 
         std::priority_queue<std::pair<dist_t, labeltype >>
-        searchKnn(const void *query_data, size_t k) const {
+        searchKnn(const void *query_data, size_t k, filter_func_t& isIdAllowed=allowAllIds) const {
             std::priority_queue<std::pair<dist_t, labeltype >> topResults;
```
dyashuni, line 96: Could you please add the same flag as in `hnswlib/hnswalg.h` and check of `k`

```
assert(k <= cur_element_count);
bool is_filter_disabled = std::is_same<filter_func_t, decltype(allowAllIds)>::value;
```

hnswlib/bruteforce.h
```diff
@@ -92,23 +92,30 @@ namespace hnswlib {
...
             if (cur_element_count == 0) return topResults;
             for (int i = 0; i < k; i++) {
                 dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
-                topResults.push(std::pair<dist_t, labeltype>(dist, *((labeltype *) (data_ + size_per_element_ * i +
-                                                                                    data_size_))));
+                labeltype label = *((labeltype*) (data_ + size_per_element_ * i + data_size_));
+                if(isIdAllowed(label)) {
```
dyashuni, line 101: `if(is_filter_disabled || isIdAllowed(label)) {`

hnswlib/bruteforce.h
```diff
@@ -92,23 +92,30 @@ namespace hnswlib {
...
             }
-            dist_t lastdist = topResults.top().first;
+            dist_t lastdist = topResults.empty() ? std::numeric_limits<dist_t>::max() : topResults.top().first;
             for (int i = k; i < cur_element_count; i++) {
                 dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
                 if (dist <= lastdist) {
-                    topResults.push(std::pair<dist_t, labeltype>(dist, *((labeltype *) (data_ + size_per_element_ * i +
-                                                                                        data_size_))));
+                    labeltype label = *((labeltype *) (data_ + size_per_element_ * i + data_size_));
+                    if(isIdAllowed(label)) {
```
dyashuni, line 110: `if (is_filter_disabled || isIdAllowed(label)) {`

dyashuni:
Thank you!

gtsoukas:
This merge is great news! Are there plans to include this in the next release? Are there plans for implementing the Python-interface for the filtering functionality?

dyashuni:
@gtsoukas Thank you! We can add filtering to Python. Just need to figure out what are the most popular use cases for filtering and create a proper API. What kind of filtering are you interested in Python? In my head we can filter elements that are

1. in the list of allowed elements - include elements that are in the list
2. not in the list of disallowed elements - include elements that are not in the list

Any other ideas?

gtsoukas:
Thank you @dyashuni! My use case is about a few millions of vectors  and it requires filtering varying fractions _f_ in sub 5ms per query, where _f_ can be a few hundred vectors but also 50% of all vectors i.e. millions. A query typically must return a few hundred nearest vectors. One example is filtering for all female clothes in an e-commerce store.

Here are some preliminary thoughts:
> 1. in the list of allowed elements - include elements that are in the list

This is probably only fast if _f_ is small, i.e. a few hundred/thousand ids. For large _f_ I guess that the time for assembling the list and potentially transform from python to C++ would have a too big negative effects on performance. For very small _f_ on the other hand, exact ANN e.g. via BLAS might be the better option anyway.

> 2. not in the list of disallowed elements - include elements that are not in the list

This would work well if _f_ is extremely large, e.g. 99%. For large _f_ on the other hand post-filtering might be OK for real world use cases i.e. query a few more elements than actually needed and delete the disallowed elements from the results. But still the proposed filter would be better than post filtering, since k is guaranteed and efficiency is better.

In summary list of allowed/disallowed elements would probably only work well for very small _f_ and very large _f_ but would not be efficient for a very large range of _f_.

Another option would be to have an optional _filter_ function in the python API which takes an id as parameter and returns a boolean. My guess is, that this would create a too large performance penalty due to the large amount of calls from C++ to python. What do you think? (I don't have any experience with implementing C++ python bindings)

Probably the best option would be to allow tagging the nodes/vectors. This could be done via an optional parameter _tags_ or _categories_ to the _add_items()_ function. Then, instead of passing a list of node ids to the _knn_query()_ function API users could add a very small list of _allowed_tag_ids_ or _disallowed_tag_ids_. This is probably the best option for efficiently handling the entire range of _f_. The largest downside is, that it is not only a change to the python API but rather requires some additional functionality to the core library. The tagging would require maintaining a vector of hash-maps. I would argue, that for many real world use cases this functionality is vital. What do you think?

These are just some preliminary thoughts. I have already started working on a pull request. If I will come up with anything useful, I will mention it here within the next 14 days. Anyway would love to hear your thoughts.

PR #404, opened 2022-08-21
Port git_tester.py on Windows

dyashuni:
Update git_tester.py script to support both Windows and Linux platforms

### DIFF
--- a/examples/git_tester.py
+++ b/examples/git_tester.py
@@ -1,34 +1,44 @@
+import os
+
+from sys import platform
 from pydriller import Repository
-import os 
-import datetime
-os.system("cp examples/speedtest.py examples/speedtest2.py") # the file has to be outside of git
-for idx, commit in enumerate(Repository('.', from_tag="v0.6.0").traverse_commits()):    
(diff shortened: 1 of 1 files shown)


examples/git_tester.py
```diff
@@ -1,34 +1,44 @@
...
+if platform == "win32":
+    copy_cmd = "copy"
+    rm_dir_cmd = "rmdir /s /q"
+else:
+    copy_cmd = "cp"
+    rm_dir_cmd = "rm -rf"
 
+speedtest_src_path = os.path.join("examples", "speedtest.py")
+speedtest_copy_path = os.path.join("examples", "speedtest2.py")
+os.system(f"{copy_cmd} {speedtest_src_path} {speedtest_copy_path}")  # the file has to be outside of git
```
yurymalkov, line 15: Maybe use `shutil.copyfile`and `shutil.rmtree` instead of checks?
I also think "/" works in windows as well.

dyashuni: Added shutil.
os.path.join is a standard way of combining paths. I'd like to keep it.

PR #410, opened 2022-09-04
Refactoring

dyashuni:
Added a new method to replace the deleted elements and tests
Refactoring:
- deleted unused function searchKnnInternal
- improved code readability

Edit:
I left only refactoring

### DIFF
--- a/python_bindings/bindings.cpp
+++ b/python_bindings/bindings.cpp
@@ -9,7 +9,7 @@
 #include <assert.h>
 
 namespace py = pybind11;
-using namespace pybind11::literals; // needed to bring in _a literal
+using namespace pybind11::literals;  // needed to bring in _a literal
 
 /*
  * replacement for the openmp '#pragma omp parallel for' directive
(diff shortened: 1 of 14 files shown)


yurymalkov:
Can you split it into 2 PRs? 
It is hard to review both refactoring code changes (which don't require attention) and meaningful changes (which require a lot of attention).

python_bindings/bindings.cpp
```diff
@@ -74,60 +74,97 @@ inline void ParallelFor(size_t start, size_t end, size_t numThreads, Function fn
...
-    if(space_name=="l2") {
-      l2space = new hnswlib::L2Space(dim);
-    }
-    else if(space_name=="ip") {
-      l2space = new hnswlib::InnerProductSpace(dim);
-    }
-    else if(space_name=="cosine") {
-      l2space = new hnswlib::InnerProductSpace(dim);
-      normalize=true;
+inline void set_input_array_shapes(const py::buffer_info& buffer, size_t* rows, size_t* features) {
```
yurymalkov, line 82: Sound like it should be get_array_shape or something.

python_bindings/bindings.cpp
```diff
@@ -74,60 +74,97 @@ inline void ParallelFor(size_t start, size_t end, size_t numThreads, Function fn
...
-    ep_added = true;
-    index_inited = false;
-    num_threads_default = std::thread::hardware_concurrency();
+}
 
-    default_ef=10;
-  }
 
-  static const int ser_version = 1; // serialization version
+inline std::vector<size_t> get_input_ids(const py::object& ids_, size_t rows) {
```
yurymalkov, line 94: Seems like this function translates ids_ into std::vector<size_t> ignoring the possible shape. It is not clear from the naming... 
Also not clear why we need `rows` here. There is a check, but I guess it should be done outside of the conversion.

yurymalkov (commented):
The PR has some simple refactoring (I guess automated or semi-automated), fixing some issues with tests and removal of some code duplicates.
The automated refactoring together is ugly diff makes it hard to spot the meaningful changes (and it does not make much sense to check the automated refactored code).
Maybe, can you provide a list of those changes, comment them or split the PR even more?

dyashuni:
I reverted changes in python tests and code duplication changes. Now only refactoring is here.
- code formatting is based on the cpplint https://github.com/cpplint/cpplint that follows Google C++ style guide. The main changes: adding and removing redundant spaces, new lines, remove indents within namespaces, brackets "{}" positions in if/else statements
- changing declaration order: data members first
- remove unused function searchKnnInternal

python_bindings/bindings.cpp
dyashuni: Will be addressed in next PR

python_bindings/bindings.cpp
dyashuni: Will be addressed in next PR

yurymalkov:
Thank you! With a quick look seems to be good. We would need to run the performance tests before the release though.

Please merge pull request with the "squash and merge" option, so it would be a single commit.

dyashuni:
Squashed changes in one commit. Performance tests do not show any regress.

Issue #294, opened 2021-03-12
Support "skipping" nodes during traversal?


gosha1128:
Is it possible to tell the HNSW algorithm to skip a given set of nodes during search/traversal?

Note that I do not want "delete" nodes.

I just want the traversal to skip a set of pre-determined nodes when it's deciding to add any of those "skip nodes" to list of nearest neighbors, as its traversing the graph.

I think this means that for a "Top-K" list of nearest neighbors - the kind of search I describe may have to perform a few more traversals to get to K items ( since it may have "skipped" a few nodes along the way. )

I don't want "delete" because the "skip node list" may be different between searches on the same graph.

yurymalkov:
Hu @gosha1128,
That sure if possible, but requires alteration of the C++ code (can point out to the place in the code).

gosha1128:
Yes that would be great- thanks.

yurymalkov:
In the main search cycle https://github.com/nmslib/hnswlib/blob/master/hnswlib/hnswalg.h#L249 you can replace all `isMarkedDeleted(some_internal_id)` to a check that `some_internal_id` does not belong to this node set.

PR #411, opened 2022-09-05
add tag search

caohassl:
I notice the issue（Support "skipping" nodes during traversal? #294） and support it , further supports tag search～

1、when building the index , enable to define the tag of the vector 
2、when the build is finished，add a inverted file to store those tags 
3、when querying,  use the inverted file to skip nodes by tag (or id)
4、add a python file to test

### DIFF
--- a/hnswpf/hnsw_core.h
+++ b/hnswpf/hnsw_core.h
@@ -2,8 +2,20 @@
 #ifndef HNSWPF_CORE_H_
 #define HNSWPF_CORE_H_
 
-#include <dirent.h>
+#if defined(_WIN32)
+#include <direct.h>
+#include <io.h>
+#define MKDIR(a) _mkdir((a))
+#define FILE_SEPARATOR "\\"
+#else
+
 #include <ftw.h>
+
+#define FILE_SEPARATOR "/"
+#define MKDIR(a) mkdir((a),0755)
+
+#endif
+
 #include "hnswlib.h"
 #include "hnswalg.h"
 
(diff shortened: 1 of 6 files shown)


yurymalkov:
Hi @caohassl ,

Thank you for PR! It seems it has some conflict (in terms of functionality) with #402, so we would need to figure out how to merge them. 
It also seems that windows tests are failing due to `dirent.h`

caohassl:
Hi @yurymalkov 

Thanks for the reply！
I have read the PR #402，it seems that the ID positive filtering(keeping the nodes we need) has already been supported .

The first scenario is that we need to discard some unnecessary nodes，not only just keep the nodes we need in the scan. 
The second scenario is that we also need to find the TOPN that satisfies the particular tag, rather than taking TOPN + 1000 and filtering out 1000

To support ID filtering(positive and negative)  and tag filtering，I commit the PR (and fix the ci error)。

dyashuni:
Hi @caohassl, why did you close the PR?

caohassl:
hi @dyashuni 

I realized that a lot of filtering may cause performance problem, so I closed the PR.
I will try to add a threshold to jump out the scan earlier, if the performance is good, and then I will reopen it

dyashuni:
@caohassl Thank you. Yes, heavy filtering makes search slow.
I think there is no need to create a new class `SearchHNSW` because its code is almost the same as the code of the `HierarchicalNSW` class. You can use the `HierarchicalNSW` class with a custom implementation of `FilterFunctor` to avoid code duplication. If serialization is required you can add serialization/deserialization methods to your implementation of `FilterFunctor` as well.

PR #414, opened 2022-09-17
Python recall test update

dyashuni:
Added -v to python unittest discovery to get nicely looking logs in actions (we can see the test names and results):

```
testRandomSelf (bindings_test.RandomSelfTestCase) ... ok
testGettingItems (bindings_test_getdata.RandomSelfTestCase) ... ok
...
```

Update tests in bindings_test_recall.py. Previously the tests almost always passed because there were no checks for recall values. I added a threshold to check that recall is high enough and doesn't change after index serialization.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,15 +3,15 @@ name: HNSW CI
 on: [push, pull_request]
 
 jobs:
-  test:
+  test_python:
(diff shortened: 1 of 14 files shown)


dyashuni:
I messed history up. I will recreate the branch and PR

PR #418, opened 2022-09-20
Replace deleted elements at addition

dyashuni:
Added a new method to replace a deleted element when a new one is added:
`labeltype addPointToVacantPlace(const void* data_point, labeltype label)`

The method is the same as `addPoint` but allows to replace a deleted element with a new one to save memory. Note: addPointToVacantPlace is much slower than addPoint, because replacement is a heavy operation.

Edit:
- Added replace a deleted element when a new one is added
- Fix of https://github.com/nmslib/hnswlib/issues/420 by introducing locks by label
- Added a multi thread test to test against concurrent operations with the same labels
- Added timeout to test jobs in actions in case if we have deadlock
- Refactoring

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,7 @@ jobs:
           ./searchKnnCloserFirst_test
           ./searchKnnWithFilter_test
           ./multiThreadLoad_test
+          ./multiThread_replace_test
           ./test_updates
           ./test_updates update
         shell: bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6.15", "3.7", "3.8", "3.9", "3.10"]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,6 +59,7 @@ jobs:
           fi
           ./searchKnnCloserFirst_test
           ./searchKnnWithFilter_test
+          ./multiThreadLoad_test
           ./test_updates
           ./test_updates update
         shell: bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.6.15", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,7 @@ jobs:
         run: python -m pip install .
       
       - name: Test
+        timeout-minutes: 15
         run: python -m unittest discover -v --start-directory python_bindings/tests --pattern "*_test*.py"
   
   test_cpp:
@@ -52,6 +53,7 @@ jobs:
         shell: bash
       
       - name: Test
+        timeout-minutes: 15
         run: |
           cd build
           if [ "$RUNNER_OS" == "Windows" ]; then
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -451,6 +451,8 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
         tableint next_closest_entry_point = selectedNeighbors.back();
 
         {
+            // lock only during the update
+            // because during the addition the lock for cur_c is already acquired
             std::unique_lock <std::mutex> lock(link_list_locks_[cur_c], std::defer_lock);
             if (isUpdate) {
                 lock.lock();
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -16,7 +16,8 @@ typedef unsigned int linklistsizeint;
 template<typename dist_t>
 class HierarchicalNSW : public AlgorithmInterface<dist_t> {
  public:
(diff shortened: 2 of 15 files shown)


hnswlib/hnswalg.h
```diff
@@ -800,6 +812,48 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t, filter_func_t> {
     }
 
 
+    /**
+    * Adds point and replaces previously deleted point if any, updating it with new point
+    *
+    * If deleted point was replaced returns its label, else returns label of added point
+    */
+    labeltype insertPoint(const void* data_point, labeltype label) {
```
yurymalkov, line 820: I think the naming is confusing. Maybe `addToVacantPlace` of something...

hnswlib/hnswalg.h
```diff
@@ -800,6 +812,48 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t, filter_func_t> {
...
+        if (!replace_deleted) {
+            throw std::runtime_error("Don't use insertPoint when replacement of deleted elements is disabled");
```
yurymalkov, line 822:
```suggestion
            throw std::runtime_error("Can't use insertPoint when replacement of deleted elements is disabled");
```

hnswlib/hnswalg.h
```diff
@@ -800,6 +812,48 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t, filter_func_t> {
...
+        std::unique_lock <std::mutex> tmp_del_el_lock(deleted_elements_lock);
+        bool is_empty = deleted_elements.empty();
+        tmp_del_el_lock.unlock();
+
+        if (is_empty) {
+            addPoint(data_point, label);
+            return label;
+        }
+        else {
+            tmp_del_el_lock.lock();
```
yurymalkov, line 834: It will crash here with parallel additions as another thread can access it between the locks.

yurymalkov (commented):
It looks like the code is not thread safe.

I guess we should write parallel stress tests to find those bugs.

hnswlib/hnswalg.h
```diff
@@ -799,10 +842,64 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
-        * Adds point. Updates the point if it is already in the index
+    /*
+    * Adds point and replaces previously deleted point if any, updating it with new point
+    * If deleted point was replaced returns its label, else returns label of added or updated point
+    * 
+    * Note:
+    *  Methods that can work with deleted elements unmarkDelete and addPoint are not safe to use
+    *  with this method, because addPointToVacantPlace removes deleted elements from the index. 
+    */
+    labeltype addPointToVacantPlace(const void* data_point, labeltype label) {
```
yurymalkov, line 853: Can we make it a flag to `addPoint` instead of a separate function? 
I guess users who opt to use the replacement might still want to use fast insertions in some cases.  
A flag might cause to break brute force, but I think we can throw an exception that it is unsupported.

yurymalkov (changes requested):
Thank you! 
I think the main thing to consider is the interface - using flag instead of separate method.

hnswlib/hnswalg.h
```diff
@@ -57,7 +54,8 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
 
     DISTFUNC<dist_t> fstdistfunc_;
     void *dist_func_param_{nullptr};
-    std::mutex label_lookup_lock;
+
+    mutable std::mutex label_lookup_lock;  // lock for label_lookup_
```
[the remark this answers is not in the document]
dyashuni: we use mutex in const function `getDataByLabel`, so we must write `mutable`
https://cplusplus.com/forum/beginner/219016/

hnswlib/hnswalg.h
```diff
@@ -799,11 +842,47 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
-    /**
-        * Adds point. Updates the point if it is already in the index
+    /*
+    * Adds point. Updates the point if it is already in the index.
+    * If replacement of deleted elements is enabled: replaces previously deleted point if any, updating it with new point
     */
-    void addPoint(const void *data_point, labeltype label) {
-        addPoint(data_point, label, -1);
+    void addPoint(const void *data_point, labeltype label, bool replace_deleted = false) {
+        if ((replace_deleted_ == false) && (replace_deleted == true)) {
```
yurymalkov, line 850: Maybe rename `replace_deleted_` to `allow_replace_deleted_` so those are not confused?

yurymalkov (commented):
I think the parameter in the index should be renamed as there is a name collision. Maybe also run the performance regression tests. Otherwise looks great to me! Thank you!

.github/workflows/build.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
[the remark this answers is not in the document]
dyashuni: there is no available build of Python 3.6 for ubuntu-22.04

https://github.com/actions/setup-python/issues/544

hnswlib/hnswalg.h
dyashuni: Good point

hnswlib/hnswalg.h
```diff
@@ -451,6 +451,10 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
         tableint next_closest_entry_point = selectedNeighbors.back();
 
         {
+            std::unique_lock <std::mutex> lock(link_list_locks_[cur_c], std::defer_lock);
```
yurymalkov, line 454: I wonder, what is the purpose of `std::defer_lock`? Are there deadlocks without it?

dyashuni: without defer_lock we have a dead lock in `tableint addPoint(...)` method because there `mutuallyConnectNewElement` is called under the locking the `link_list_locks_[cur_c])`. So we have two locks of `cur_c`.
I lock the `cur_c` only if element is updated. it helps to avoid dead lock when element is added.

yurymalkov (approved):

PR #427, opened 2022-12-05
Add HierarchicalNSW::indexFileSize() function for precise memory footprint control

drons:


yurymalkov:
Hi @drons,
Thank for the PR! We will check it and merge it after we prepare the release (should happen soon).

python_bindings/bindings.cpp
```diff
@@ -218,6 +218,9 @@ class Index {
         this->num_threads_default = num_threads;
     }
 
+    void indexFileSize() const {
```
dyashuni, line 221: we need to return an index size here
```cpp
size_t indexFileSize() const {
    return appr_alg->indexFileSize();
}
```

dyashuni (changes requested):
Thank you, please add a small change

python_bindings/bindings.cpp
drons: :man_facepalming:

yurymalkov (approved):

yurymalkov:
Thank you for the PR!

drons:
Thanks!

PR #430, opened 2023-01-12
Fix insufficient results during filtering

kishorenc:
Very similar in nature to https://github.com/nmslib/hnswlib/pull/344

I've discussed this fix with @dyashuni before and this has been tested on a number of datasets.


dyashuni:
Thank you, @kishorenc 

Idea of the fix:
if filter is enabled (`!isIdAllowed` is `false`) we don’t care about performance and we should check 
`top_candidates.size() == ef`

if filter is disabled (`!isIdAllowed` is `true`) we should have the same behaviour as without filter
`top_candidates.size() == ef || !has_deletions`

dyashuni:
Thank you, @kishorenc

PR #431, opened 2023-01-12
Getters for max elements, element count and num deleted.

kishorenc:
These getters are useful for resizing of the index from outside when the element count reaches capacity.


dyashuni (approved):

dyashuni:
Thank you, @kishorenc

PR #432, opened 2023-01-14
Refactoring of project structure

dyashuni:
- Refactoring of project structure
   - split tests to cpp and python
   - keep in `examples` folder only examples
   - move tests to separate folder
- Add page with examples and new example files
- Update Readme


yurymalkov (approved):
Overall looks good to me. There is a misprint in readme's code and few suggestions.

PR #433, opened 2023-01-14
Add code comments that python filter works slow in multi-threaded mode

dyashuni:
Add warning that python filter works slow in multi-threaded mode

Edit:

- added comments with warnings in examples
- added examples to CI test


yurymalkov:
I can imagine a scenario where a user is overwhelmed with those warning messages. 
I guess we can put a condition that we show it only once index and/or we can add log levels, though that would require some designing.

PR #446, opened 2023-03-19
Add CMake install targets

moritz-h:
Add CMake install targets to allow installing the library using CMake and allow usage with find_package().

Other changes:
- Raise the minimum CMake version to 3.0. Testing CMake 2.8.12.2 on CentOS 7 generated errors already before doing these changes.
- Make the CMake project() name consistent with the library name (change `hnsw_lib` to `hnswlib`).
- Add option `HNSWLIB_EXAMPLES` to allow enabling/disabling tests and examples explicitly. Option defaults to previous behavior checking `CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.0...3.26)
 
 project(hnswlib
     LANGUAGES CXX)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,36 @@
-cmake_minimum_required (VERSION 2.6)
-project(hnsw_lib
+cmake_minimum_required(VERSION 3.0)
+
+project(hnswlib
     LANGUAGES CXX)
 
-add_library(hnswlib INTERFACE)
-target_include_directories(hnswlib INTERFACE .) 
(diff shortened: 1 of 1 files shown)


moritz-h:
May I ask is there a tendency if these changes will be ok or not?
Installing CMake targets for hnswlib would make packaging of downstream C++ libraries with a transitive dependency on hnswlib a lot easier. But if you don't want to change CMake here, I would at least know that downstream workarounds are the way to go in the long run.

yurymalkov:
Hi @moritz-h,
Thank you for the PR and the reminder! 
I'm open to merging it, but as I'm not very familiar with CMake, I'm a bit concerned about potential negative consequences for users relying on the previous structure. What do you think are the chances of a broken build in such cases?

moritz-h:
I just discovered that the next CMake release 3.27 will print a deprecation warning for projects with compatibility to CMake older than 3.5, see https://cmake.org/cmake/help/git-master/release/dev/deprecate-policy-old.html
This is fixed by setting a version range to cmake_minimum_required to explicitly set the project compatible to newer CMake versions.

yurymalkov:
Cool, thank you! We will merge the PR after @dyashuni will do a quick look at it.

dyashuni:
@moritz-h Thank you, looks good

PR #363, opened 2022-01-27
Disable march=native option when building on Apple M1

yoshoku:
I failed to install hnswlib with pip on Apple M1. The output error message is as follows:

```sh
$ pip install hnswlib
Collecting hnswlib
  Using cached hnswlib-0.6.0.tar.gz (30 kB)

...

Building wheels for collected packages: hnswlib
  Building wheel for hnswlib (pyproject.toml) ... error
  ERROR: Command errored out with exit status 1:

...

  clang: error: the clang compiler does not support '-march=native'
  error: command '/usr/bin/clang' failed with exit code 1
  ----------------------------------------
  ERROR: Failed building wheel for hnswlib
Failed to build hnswlib
ERROR: Could not build wheels for hnswlib, which is required to install pyproject.toml-based projects
$
```

The cause of this error is that Apple M1's clang does not support the `march=native` option.  I have fixed setup.py so that the `march=native` option is not given when building on Apple M1. With this fix, I have confirmed that the installation is successful:

```sh
$ pip install .
Processing /Users/yoshoku/hnswlib

...

Building wheels for collected packages: hnswlib
  Building wheel for hnswlib (pyproject.toml) ... done

...

Successfully built hnswlib
Installing collected packages: hnswlib
Successfully installed hnswlib-0.6.0
$
```

PR #455, opened 2023-04-22
Resolve wheel build failure on x86_64 architecture due to -march=native flag

XiChenn:
The hnswlib wheel build failed on my MacOS (x86_64 architecture) when I run ```pip install hnswlib```. 
```
      clang: error: the clang compiler does not support '-march=native'
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for hnswlib
Failed to build hnswlib
ERROR: Could not build wheels for hnswlib, which is required to install pyproject.toml-based projects
```
Similar to the solution of the previous merged [PR](https://github.com/nmslib/hnswlib/pull/363), by removing the building flag resolves the issue.
```
pip install .
Looking in indexes: https://pypi.python.org/simple/
...
Successfully built hnswlib
Installing collected packages: hnswlib
Successfully installed hnswlib-0.7.0

```


yurymalkov:
Hm. I can imagine this would remove native optimizations from linux builds. 
Do I understand what is causing the problem?

yurymalkov:
We can try to understand if the compiler is clang with something like 

```python
import sys
import platform
import subprocess
import re

def is_clang_compiler():
    try:
        # Run the compiler version command
        result = subprocess.run(['cc', '--version'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
        output = result.stdout.lower()

        # Check if the output contains the word 'clang'
        return 'clang' in output
    except FileNotFoundError:
        # If the 'cc' command is not found, it's not Clang
        return False

**Comment by yurymalkov** [comment]
Can you double check that your compiler is set up properly? I've added macos to the CI with x64 and the tests are passing.
I am also confused that your error message has both gcc and clang mentioned, sounds fishy

**Comment by XiChenn** [comment]
My Mac is 2019 version.
see the result below:
```bash
>>> result = subprocess.run(['cc', '--version'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
>>> output = result.stdout.lower()
>>> output
'apple clang version 14.0.3 (clang-1403.0.22.14.1)\ntarget: x86_64-apple-darwin22.4.0\nthread model: posix\ninstalleddir: /applications/xcode.app/contents/developer/toolchains/xcodedefault.xctoolchain/usr/bin\n'
```

dyashuni:
@XiChenn Could you please try `pip install .` from PR https://github.com/nmslib/hnswlib/pull/461 ?
it should fix your issue

yurymalkov:
@XiChenn Can you check if the other change fixes your problem by running ```pip install git+https://github.com/nmslib/hnswlib.git@refs/pull/461/head```?

cterrazas2:
fwiw, @yurymalkov I was having the same issue as @XiChenn (Mac x86_64 architecture) and installing via #461's head worked for me.

Issue #107, opened 2019-05-28
heap-buffer-overflow when sse enabled


supercoolgreatcoder:
Steps to reproduce:
1. Build with Address Sanitizer enabled (add -fsanitize=address -fno-omit-frame-pointer flags to compiler and linker):
https://gist.github.com/supercoolgreatcoder/11b5deccd1f500bd19c84cdf11f2c2a9
2. run sift1B test:
https://gist.github.com/supercoolgreatcoder/764ca0f196bee8a28b629e9e3c56f945
Culprit line: https://github.com/nmslib/hnswlib/blob/d0716d9962ec4ba100a5de551f07c2067e835785/hnswlib/hnswalg.h#L199
Disabling SSE does not trigger heap buffer overflow.

yurymalkov:
Hi @supercoolgreatcoder,
Thank you! What sort of problems it might create?
Actually, I thought `_mm_prefetch` is immune to wrong addresses.

supercoolgreatcoder:
@yurymalkov, __mm_prefetch is trying to read chunk of memory in heap that might not have been allocated, e.g. `int* a = new int[100]; std::cout << a[100]`. In the worst case it might end up with a segfault. I believe, it should be fixed relatively easy by fixing number of bytes that has to be prefetched.

fuzesmarcell:
This issue is still persisting and the root cause is the following expression: `*(datal + j + 1)`

The issue is that for `datal` we already account for the size at the beginning of the buffer here:
```
                size_t size = getListCount((linklistsizeint*)data);
                tableint *datal = (tableint *) (data + 1);
```

So adding the extra `1` in the loop is incorrect and on the last iteration we are de-referencing a out of bounds value.
The following fixes the issue and does not cause any heap buffer overflow
```
                    _mm_prefetch((char *) (visited_array + *(datal + j)), _MM_HINT_T0);
                    _mm_prefetch(getDataByInternalId(*(datal + j)), _MM_HINT_T0);
```

PR #459, opened 2023-05-05
Fix heap buffer overflow caused by prefetch.

kishorenc:
The prefetching goes past `size` since `datal` is initialized 1 index past `data`. 

I uncovered this issue on ASAN when SSE is enabled. Also spotted earlier here: https://github.com/nmslib/hnswlib/issues/107

cc @dyashuni

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -249,8 +249,8 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
                 tableint candidate_id = *(datal + j);
 //                    if (candidate_id == 0) continue;
 #ifdef USE_SSE
-                _mm_prefetch((char *) (visited_array + *(datal + j + 1)), _MM_HINT_T0);
-                _mm_prefetch(getDataByInternalId(*(datal + j + 1)), _MM_HINT_T0);
+                _mm_prefetch((char *) (visited_array + *(datal + j)), _MM_HINT_T0);
+                _mm_prefetch(getDataByInternalId(*(datal + j)), _MM_HINT_T0);
 #endif
                 if (visited_array[candidate_id] == visited_array_tag) continue;
                 visited_array[candidate_id] = visited_array_tag;


hnswlib/hnswalg.h
```diff
@@ -260,8 +260,8 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
                 tableint candidate_id = *(datal + j);
 //                    if (candidate_id == 0) continue;
 #ifdef USE_SSE
-                _mm_prefetch((char *) (visited_array + *(datal + j + 1)), _MM_HINT_T0);
-                _mm_prefetch(getDataByInternalId(*(datal + j + 1)), _MM_HINT_T0);
+                _mm_prefetch((char *) (visited_array + *(datal + j)), _MM_HINT_T0);
+                _mm_prefetch(getDataByInternalId(*(datal + j)), _MM_HINT_T0);
```
yurymalkov, line 264: I think that fix defeats the purpose of the prefetch, as it need to be done for the next element.
Is there any evidence this prefetch affects the program behavior? 
If so, I guess we probably should come with a different fix, e.g. min(j+1, max_value)

kishorenc: Given that a cache line is atleast 64 bytes, does the extra `+1` really matter? What matters is that the loop is moving forward via the `j` variable increments so we keep fetching ahead. We should be able to confirm this via a benchmark: I don't think there will be any degradation in performance.

There are actually a couple of other places the same issue happens, which I just noticed: 

- https://github.com/nmslib/hnswlib/blob/master/hnswlib/hnswalg.h#L346
- https://github.com/nmslib/hnswlib/blob/master/hnswlib/hnswalg.h#L1010

kishorenc: Also, I think the proper way to actually use a `_mm_prefetch` is to fetch a full 64-bytes ahead, for example how it's done here by fetching "8 doubles ahead": 

https://stackoverflow.com/questions/73012072/why-such-address-is-used-with-mm-prefetch

yurymalkov:
```C++ 
_mm_prefetch((char *) (visited_array + *(datal + j + 1)), _MM_HINT_T0);
```

+1 fetches the visited bit for the next item, which can be anywhere in the memory compared the previous element. Almost certain different cache and page and at least skylake's cpus were struggling without the prefetches, that is why it ended up in the code despite the obvious issue.

Yeah, it would be great if you could profile the performance on few different x64 platforms and maybe improve the prefetching performance and hopefully fix this out-of-bound issue that made people uneasy. 

I think the proper way to test the performance is to do search on small dimensional data to focus on memory sparsity (I used d=4)  and keep two implementations of the search function (with and without the change), so the performance would not depend on random memory layout changes in the data. That should allow measuring the performance within 1-2% accuracy.

kishorenc:
@yurymalkov 

> +1 fetches the visited bit for the next item, which can be anywhere in the memory compared the previous element. 

Ah, sorry, I missed the extra pointer de-referencing in the code and assumed a sequential lookup.

I benchmarked with/without prefetching by searching on 1M / 4-dimension points.

- with prefetching: 6700 ms
- without prefetching: 7800 ms

So, I've now updated the PR to use `std::min()` to limit the index of the next prefetch. This is comparable in performance to the current version without bound checking.

yurymalkov:
Hi @kishorenc,
Thanks for the change! We would need to double check the performance

yurymalkov:
@kishorenc can you please share the benchmarking code? 
Also, I wonder, how do you look at the other `*(data + 1)` in the code? Those also have the same out-of-bounds prefetch.

kishorenc:
Here's [the code for the benchmark](https://gist.github.com/kishorenc/deb64aa090bcc9eb2d4a16d4cb27d2d6). I just ran this with/without the changes in the PR and compared.

> Also, I wonder, how do you look at the other *(data + 1) in the code? Those also have the same out-of-bounds prefetch.

Yes, those are out-of-bound as well, but I didn't address them because I was not sure what their actual sizes were. In the places that I've changed, there was a clear `size` variable available that I could go by.

czpmango (approved):

Issue #318, opened 2021-06-03
assert in checkIntegrity


gozyw:
line 1168：assert(data[j] > 0); 
is it uncorrect? i think it should be data[j] >= 0

dyashuni:
will be fixed here
https://github.com/nmslib/hnswlib/pull/461

PR #461, opened 2023-05-10
Fix mac setup

dyashuni:
- add macos to CI
- allow asserts in C++ tests (remove -DNDEBUG)
- fix checkIntegrity data[j] is always >= 0
- add checks for compiler flags `-march=native` and `-mcpu=apple-m1` for python installation for macos
- fix compiler flags for Windows


dyashuni:
fixes
https://github.com/nmslib/hnswlib/issues/318

PR #463, opened 2023-05-17
Fix global linkage again

alxvth:
An older pull request ([#383](https://github.com/nmslib/hnswlib/pull/383)) fixed potential global linkage problems by making header-defined functions static. During some merging one of those `static`s must have been lost (looking at the commit history of `hnswlib.h` that was probably my fault [here](https://github.com/nmslib/hnswlib/commit/23f53517a82978defdec04e406b391c38222a538)).

In this PR I'm simply adding this lost `static` again.


yurymalkov:
Thank you!

PR #466, opened 2023-05-27
Fixed linking error with Visual Studio

jrade:
If you include the header hnswlib/hnswlib.h in multiple C++ files and compile with Visual Studio then you get the following error

  error LNK2005: "void __cdecl cpuid(int * const,int,int)" (?cpuid@@YAXQEAHHH@Z) already defined in Pch.obj
  fatal error LNK1169: one or more multiply defined symbols found

This is easy to fix by declaring the function cpuid() static

### DIFF
--- a/hnswlib/hnswlib.h
+++ b/hnswlib/hnswlib.h
@@ -15,7 +15,7 @@
 #ifdef _MSC_VER
 #include <intrin.h>
 #include <stdexcept>
-void cpuid(int32_t out[4], int32_t eax, int32_t ecx) {
+static void cpuid(int32_t out[4], int32_t eax, int32_t ecx) {
     __cpuidex(out, eax, ecx);
 }
 static __int64 xgetbv(unsigned int x) {


jrade:
I used the keyword static since several other functions are declared static for the same reason. But I think it would be better to declare all these functions inline instead. Why?

It is all about the ODR, the one definition rule. It says that a function can only be defined once. If you define a function in a header and include that header in 5 source files, then you have defined it 5 times which violates the ODR.

This is usually handled by declaring the function inline. Then you tell the compiler that all 5 definitions are identical definitions of one and the the same function, which is what you want here. Note that the keyword inline does not have anything to do with inlining in modern C++. It is all about the ODR. (See https://en.cppreference.com/w/cpp/language/inline)

If you handle this by declaring the function static, then you tell the compiler these are 5 different functions that just happen to have the same name, which is not what you want here. The compiler will generate 5 different function bodies. The linker will probably notice that they are identical and throw away all but one of them. This probably works but is inefficient and a bit obscure.

yurymalkov:
Hi @jrade,
Thank you for the PR and the comment. The PR in its current form seems to be a duplicate of https://github.com/nmslib/hnswlib/pull/463 which we planned to merge.

I wonder, in case we switch it to inline instead of static, is there a (probably very small) chance of cpuid signature collision with  other libraries?

jrade:
There is such a risk regardless of whether the function is declared inline or static. Is it possible to move the functions at the beginning of hnswlib.h inside the namespace hnswlib?

jrade:
I made another commit where I changed all static functions to inline and moved all functions inside namespace hnswlib

yurymalkov:
Hey @jrade,

It looks like tests are not passing with 

```D:\a\hnswlib\hnswlib\hnswlib\space_l2.h(147): error C7525: inline variables require at least '/std:c++17'```

Is it a part of C++17?

jrade:
That's right. I have reverted those lines, so now it should be C++11 only.

yurymalkov:
Hi @jrade. It still does not pass the tests. Can you please look into it?

jrade:
Missed one line. Fixed now.

yurymalkov:
Got it, thank you!

PR #470, opened 2023-06-17
Simplify VisitedList.

ttsugriy:
It's error-prone to use raw pointers instead of `std::unique_ptr` which has no extra overhead compared to raw pointer but is easier to use. Also, it's better to use member initializer instead of constructor body.

### DIFF
--- a/hnswlib/visited_list_pool.h
+++ b/hnswlib/visited_list_pool.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <memory>
 #include <mutex>
 #include <string.h>
 #include <deque>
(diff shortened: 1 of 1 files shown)


yurymalkov:
Hi @ttsugriy , thank you for the pull request! 
It looks like it causes some errors in the CI. Can you please look at it?
I also wonder if you tested that it does not affect the performance?

ttsugriy:
Hello, Yury! Unfortunately it looks like there are many callsites that use explicit pointer type instead of `auto`, so many more changes are required to migrate to unique_ptr. As for performance, this change doesn't affect it, since member initialization change involves only primitive types and unique_ptr is identical to raw pointer in the assembly.

yurymalkov:
Got it, thanks!

ttsugriy:
Thank you, @yurymalkov! I'm a huge fan of HNSW, so thank you very much for inventing it!

yurymalkov:
@ttsugriy Thank you for the kind words! A lot of the hnswlib's value came from external PR contributions! :)

PR #471, opened 2023-06-19
Fix build warnings.

ttsugriy:
`sprintf` is deprecated and not secure, since it doesn't protect against buffer overflows. Also, string literals are `const char*`, not `char*`.


yurymalkov:
Thanks!

PR #472, opened 2023-06-19
Replace priority_queue::push with emplace.

ttsugriy:
It's more concise and more importantly forwards args so that the pair can be constructed in-place, instead of being created at the call-site and moved into push. It's not a big deal as both values are primitive types, but still...


yurymalkov:
Thanks!

PR #473, opened 2023-06-19
[bruteforce] Fix bruteforce removePoint.

ttsugriy:
Previous implementation was not thread-safe - unlike `addPoint` it didn't acquire an index lock.

While at it, this change also switches to `.find()` for cur_c lookups, so that following `erase` wouldn't have to perform another lookup.

P.S.: I don't see how `addPoint` is thread-safe as well, since write to data_ is performed outside of critical section, but oh well...

### DIFF
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -84,10 +84,16 @@ class BruteforceSearch : public AlgorithmInterface<dist_t> {
 
 
     void removePoint(labeltype cur_external) {
-        size_t cur_c = dict_external_to_internal[cur_external];
+        std::unique_lock<std::mutex> lock(index_lock);
 
-        dict_external_to_internal.erase(cur_external);
+        auto found = dict_external_to_internal.find(cur_external);
+        if (found == dict_external_to_internal.end()) {
+            return;
+        }
 
+        dict_external_to_internal.erase(found);
+
+        size_t cur_c = found->second;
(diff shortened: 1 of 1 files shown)


yurymalkov:
@ttsugriy 
Thank you!
Concerning the writing to `data_` - agree. We probably need to move it inside.

ttsugriy:
sounds good, @yurymalkov. Btw, unrelated to this change - have you considered using clangformat? Current formatting is somewhat inconsistent and makes reading code a bit harder. Also, if perf is important, I'd recommend considering moving away from interfaces to templates. Using interfaces for distance calculation and other strategy hooks prevents inlining, which even on its own adds overhead and branch mispredictions, but more importantly it prevents all useful compiler optimizations that rely on inlining.

yurymalkov:
Hi @ttsugriy,
I am not into strict formatting myself. I think @dyashuni did some formatting some time ago, maybe he has opinions on that.

As for interfaces/function pointers/templates I think I've tested inlining the distance computation manually long time ago and, as far as I recall, there was little if any penalty even for small dimensionalities. There is also a problem that there are several distance implementations depending on the vector length, if those are added as templates, I think the compile time will probably blow up.

yurymalkov:
Got it! I can check if it affects the performance now. 
The compile time now matters for python, as every time people do `pip install` it compiles from scratch. We can fix it with pre-compiled binaries at some point though

PR #474, opened 2023-06-19
Use unique_ptr to manage visited_list_pool_

ttsugriy:
Using raw pointers is error-prone. For example, previous implementation would leak pool in case it's created and loadIndex is invoked.


yurymalkov:
Hi @ttsugriy, it looks like the CI is not passing. Can you please look into that?

ttsugriy:
Sure thing, @yurymalkov, it's due to a difference between macOS and Linux headers. Should be fixed now.

yurymalkov:
Thanks!

PR #475, opened 2023-06-20
Efficient AVX512  implementation in 'InnerProductSIMD16ExtAVX512' Function

aurora327:
InnerProductSIMD16ExtAVX512 functions are implemented using the more efficient AVX512 instruction set

### DIFF
--- a/hnswlib/space_ip.h
+++ b/hnswlib/space_ip.h
@@ -157,19 +157,35 @@ InnerProductSIMD16ExtAVX512(const void *pVect1v, const void *pVect2v, const void
 
     __m512 sum512 = _mm512_set1_ps(0);
 
-    while (pVect1 < pEnd1) {
-        //_mm_prefetch((char*)(pVect2 + 16), _MM_HINT_T0);
-
+    size_t loop = qty16 / 4;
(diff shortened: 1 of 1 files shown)


aurora327:
Hi @yurymalko, Can you please to review the code?

yurymalkov:
Hi @aurora327,

Thank you for the PR! I am slow to respond currently due to sickness, sorry.

I wonder, how much improvement did you see in the tests with the better implementation?

aurora327:
hi, @yurymalkov 
Different  Size of vector1 and vector2 of the passed parameters have different performance gains, my own dataset build on 4th Generation Intel® Xeon® resulted in a 2% to 10% end-to-end improvement with 4 cores bound.
I hope you feel better soon :)

yurymalkov:
Thanks again for the PR! I've also checked the query performance, it is up to 15% for 16-dim.
I also wonder if aligned/unaligned memory makes a difference for the current architectures?

aurora327:
No differences were observed since the avx512 instruction used handles both aligned and unaligned data well. At the same time, the aligned buffer is highly recommended if possible.

PR #476, opened 2023-06-23
Check before removing march=native

xhochy:
If you set `HNSWLIB_NO_NATIVE` and build on Apple M1 the removal of `march=native` fails because it isn't in the list of arguments.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ class BuildExt(build_ext):
     }
 
     if sys.platform == 'darwin':
-        if platform.machine() == 'arm64':
+        if platform.machine() == 'arm64' and '-march=native' in c_opts['unix']:
             c_opts['unix'].remove('-march=native')
         c_opts['unix'] += ['-stdlib=libc++', '-mmacosx-version-min=10.7']
         link_opts['unix'] += ['-stdlib=libc++', '-mmacosx-version-min=10.7']


yurymalkov:
Hi @xhochy,
Can you check if https://github.com/nmslib/hnswlib/pull/461 resolved the issue? It is merged to `develop` branch

xhochy:
From my understanding this is even worse as it removes the option overall to omit `-march=native` from the build.

dyashuni:
@xhochy 
#461 should fix the issues with `-march=native`. It automatically checks if the flag is available on the platform. If the flag is not available it is removed. No need to set `HNSWLIB_NO_NATIVE` for macOS.
Do you think that we still need `HNSWLIB_NO_NATIVE` to manually control `-march=native` on macOS ?

xhochy:
We need a general switch to disable march=native as otherwise the binaries are no longer redistributable. The linked PR removes that possibility completely. The PR here only fixed this option for the corner case of Apple Silicon.

yurymalkov:
>We need a general switch to disable march=native as otherwise the binaries are **no longer redistributable**

@xhochy  Can you provide more details about this problem? 
Context: at some point we implemented runtime checks for AVX and AVX512 (https://github.com/nmslib/hnswlib/commit/b8010809949ae5a021997ed0e8906c319582bb0d), so for x64 I expected it to run on different (older) machines without problems.

PR #477, opened 2023-07-02
Fix memory leak on loadIndex with non-empty HierarchicalNSW object

drons:


drons:
Rebase on top of develop branch,

Part of this leak was fixed in #474.

Issue #467, opened 2023-06-15
double free bug in init_index


D4rkD0g:
Hi, the hnswlib will crashed when init index if the parameter if too big

```python
import hnswlib

h = hnswlib.Index(space='l2', dim=1)
h.init_index(max_elements=1, ef_construction=200, M=2305843009213693951)
h.add_items([1], -1)
```

the backtrace

```
Starting program: /usr/bin/python3 df.py
[*] Failed to find objfile or not a valid file format: [Errno 2] No such file or directory: 'system-supplied DSO at 0x7ffff7fc1000'
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[*] Failed to find objfile or not a valid file format: [Errno 2] No such file or directory: '.gnu_debugdata for /usr/local/lib/python3.10/dist-packages/numpy/core/../../numpy.libs/libgfortran-040039e1.so.5.0.0'
[New Thread 0x7ffff3bff640 (LWP 1541)]
[New Thread 0x7ffff33fe640 (LWP 1542)]
[New Thread 0x7ffff0bfd640 (LWP 1543)]
[New Thread 0x7fffee3fc640 (LWP 1544)]
[New Thread 0x7fffe9bfb640 (LWP 1545)]
[New Thread 0x7fffe73fa640 (LWP 1546)]
[New Thread 0x7fffe4bf9640 (LWP 1547)]
double free or corruption (out)

Thread 1 "python3" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0x0, signo=0x6, threadid=0x7ffff7c4f1c0) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
[ Legend: Modified register | Code | Heap | Stack | String ]
───────────────────────────────────────────────────────────────────────────────────────────────────────────── registers ────
$rax   : 0x0
$rbx   : 0x007ffff7c4f1c0  →  0x007ffff7c4f1c0  →  [loop detected]
$rcx   : 0x007ffff7ce6a7c  →  <pthread_kill+300> mov r13d, eax
$rdx   : 0x6
$rsp   : 0x007fffffffd4e0  →  0x007fffffffd680  →  0x007ffff757f680  →  0x00000000000012fd
$rbp   : 0x602
$rsi   : 0x602
$rdi   : 0x602
[...]

PR #484, opened 2023-07-18
hnswalg.h: cap M to 100000

emollier:
This patch works around issue #467, also referenced as CVE-2023-37365, by implementing Yury Malkov's suggestion about capping the M value, coding the maximum number of outgoing connections in the graph, to a reasonable enough value of the order of 100000.  For the record, the documentation indicates reasonable values for M range from 2 to 100, which are well within the cap; see ALGO_PARAMS.md.

The reproducer shown in issue #467 doesn't trigger the double free condition anymore after this change is applied, but completes successfully, although with the below warning popping up on purpose:

	warning: M parameter exceeds 100000 which may lead to adverse effects.
	         Cap to 100000 will be applied for the rest of the processing.

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -101,7 +101,13 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
         data_size_ = s->get_data_size();
         fstdistfunc_ = s->get_dist_func();
         dist_func_param_ = s->get_dist_func_param();
-        M_ = M;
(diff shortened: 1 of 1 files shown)


yurymalkov:
Thank you for the PR! 
On the second though it might make sense to limit it to a smaller number, like 10K (can you please decrease it?). I can hardly imagine someone would spend 40KB per item (40GB for 1M items).

yurymalkov:
Thanks!

PR #493, opened 2023-08-11
Bring back HNSWLIB_NO_NATIVE

dyashuni:
Brining back `HNSWLIB_NO_NATIVE` to fix issue https://github.com/nmslib/hnswlib/pull/476

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,10 +103,10 @@ class BuildExt(build_ext):
                 opts.append('-fvisibility=hidden')
             if not os.environ.get("HNSWLIB_NO_NATIVE"):
                 # check that native flag is available
-                print('checking avalability of flag:', native_flag)
-                if not has_flag(self.compiler, native_flag):
-                    print('removing unsupported compiler flag:', native_flag)
-                    opts.remove(native_flag)
+                print('checking avalability of flag:', self.native_flag)
+                if not has_flag(self.compiler, self.native_flag):
+                    print('removing unsupported compiler flag:', self.native_flag)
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -73,15 +73,19 @@ def cpp_flag(compiler):
 
 class BuildExt(build_ext):
     """A custom build extension for adding compiler-specific options."""
+    native_flag = '-march=native'
```
yurymalkov, line 76: Can you please name it `compiler_flag_native`?

setup.py
```diff
@@ -97,23 +101,23 @@ def build_extensions(self):
...
-                    if has_flag(self.compiler, m1_flag):
-                        print('adding flag:', m1_flag)
-                        opts.append(m1_flag)
-                    else:
-                        print(f'flag: {m1_flag} is not available')
-            else:
-                print(f'flag: {native_flag} is available')
+            if not os.environ.get("HNSWLIB_NO_NATIVE"):
+                # check that native flag is available
+                print('checking avalability of flag:', self.native_flag)
```
yurymalkov, line 106: I think it `BuildExt.native_flag` is better than `self.native_flag` for readability (as `self.native_flag` is never set)

setup.py
dyashuni: Done

xhochy:
@dyashuni Thank you for picking this up. I just came back from leave and it's nice to see someone took this.

yurymalkov:
Thank you!

PR #508, opened 2023-09-17
Provide a macro to override the use of std::cerr

jlmelville:
#484 introduced a call to `std::cerr` when `M` is excessively large. This makes life hard for the [R bindings](https://github.com/jlmelville/rcpphnsw) because CRAN checks for and does not like to find anything writing to stdout/stderr (these streams should be sent to the console instead). Fortunately, the Rcpp library provides alternative streams (e.g. `Rcpp::Rcout` and `Rcpp::Rcerr`).

I propose using a macro to define the stream. By default this will be `std::cerr`, so users of this library need do nothing and will notice no changes. In RcppHNSW, I will create a header file which will declare the override, e.g.:

```C++
#include <Rcpp.h>

#define HNSWLIB_ERR_OVERRIDE Rcpp::Rcerr

#include "hnswlib.h"
```

A similar arrangement works for [Annoy](https://github.com/spotify/annoy/blob/2be37c9e015544be2cf60c431f0cccc076151a2d/src/annoylib.h#L76-L82C7) and [RcppAnnoy](https://github.com/eddelbuettel/rcppannoy/blob/cdaeeb83c01dae7a08b011c8c2304d8487435967/inst/include/RcppAnnoy.h#L21-L22).

### DIFF
--- a/hnswlib/hnswalg.h
+++ b/hnswlib/hnswalg.h
@@ -105,8 +105,8 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
         if ( M <= 10000 ) {
             M_ = M;
         } else {
-            std::cerr << "warning: M parameter exceeds 10000 which may lead to adverse effects." << std::endl;
-            std::cerr << "         Cap to 10000 will be applied for the rest of the processing." << std::endl;
+            HNSWERR << "warning: M parameter exceeds 10000 which may lead to adverse effects." << std::endl;
+            HNSWERR << "         Cap to 10000 will be applied for the rest of the processing." << std::endl;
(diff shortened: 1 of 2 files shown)


yurymalkov:
Hi @jlmelville,

Thank you for the PR and sorry for the late response!

Looks good to me, but can you please add a link (it can be the link to this PR)?

Thanks!

jlmelville:
@yurymalkov do you mean add a link to this PR in the comment to explain why the macro has been created? No problem to do so, just want to make sure I understand the request.

yurymalkov:
Thank you!

PR #511, opened 2023-09-27
Resolve initialisation order warning

stephematician:
GNU compilers no longer warn with -Wall or -Wreorder that initialisation order does not match declaration order in HierarchicalHNSW


yurymalkov:
Thanks, @stephematician !

Issue #373, opened 2022-02-16
Cannot return the results in a contigious 2D array. Probably ef or M is too small


prateekpatelsc:
@yurymalkov : trying to understand when this can happen
I have an index where i have few hundred thousand elements , no deletions.
my topK is around range 100-500,, search_ef is ~400

Could you please elaborate in what scenarios can the algorithm run into such case ? what is the correct way to handle this
increasing M eads to increase in index sizes and slow search times , so i am not too inclined to go this route .

prateekpatelsc:
Is this of because stuck in some local minima where no neighbors are improving distance and the search queue is empty ?

prateekpatelsc:
Also is this somehow related to data dimensionality as well ? for example for a fixed M and search ef , ef_construction param , the chances of running into these kind of errors are more probable large high dimensional data or similar scale data but lower dimension

yurymalkov:
Hi @prateekpatelsc,

That might be connected to duplicates in the data. The duplicates are useless for search as each component can be substituted with a single element decreasing the index size and complexity (and one can control them on the client side doing a search before the insertion of a batch) and incremental graph-based approaches can suffer from them due to loss of connectivity in the graph if the size of a maximum duplicate component is much larger than the number of links M (and it is hard to check for duplicates for inner product inside the algorithm).

Can you check if it has large components of duplicates?

PR #513, opened 2023-10-13
New binding for returning List[List] instead of numpy array

Scikud:
Often it's useful to return lists instead of numpy arrays. This is especially useful when using filtering if k > num of items returned after filtering.  There may also be instances where you'd like to return noncontigous 2d arrays. This PR adds a new binding that will return a List of Lists instead of a numpy array. 

This pr should fix [contigious 2D array.](https://github.com/nmslib/hnswlib/issues/373)

### DIFF
--- a/python_bindings/bindings.cpp
+++ b/python_bindings/bindings.cpp
@@ -950,7 +950,7 @@ PYBIND11_PLUGIN(hnswlib) {
             py::arg("k") = 1,
             py::arg("num_threads") = -1,
             py::arg("filter") = py::none())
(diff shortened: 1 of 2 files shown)


dyashuni:
@Scikud Thank you for the PR!
Could you please modify the existing function `knnQuery_return_numpy` instead of creating a new one?
You can take a look at the `getData` function as an example. It has a `return_type` parameter that can be set to either `numpy` or `list`. Then we will need to rename the `knnQuery_return_numpy` function to something like `knnQuery`.

PR #514, opened 2023-10-18
Fix incorrect results in bruteforce with filter

lukaszsmolinski:
There is a bug in the `searchKnn` function of `BruteforceSearch`. When it is used with a filter, in some cases, it returns fewer items than expected. This PR should fix it.

The bug can be reproduced using the following code:
```python
import hnswlib

data = [[0], [1], [2]]
max_elements = len(data)
dim = len(data[0])

bf_index = hnswlib.BFIndex(space='l2', dim=dim)
bf_index.init_index(max_elements=max_elements)
bf_index.add_items(data, ids=range(max_elements))

filter_function = lambda id: id in [0, 2]
labels, _ = bf_index.knn_query(data[0], k=2, filter=filter_function)

print(labels)
# Should be: [[0 2]]
# Is: [[0 0]]
```

### DIFF
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -107,27 +107,17 @@ class BruteforceSearch : public AlgorithmInterface<dist_t> {
     searchKnn(const void *query_data, size_t k, BaseFilterFunctor* isIdAllowed = nullptr) const {
         assert(k <= cur_element_count);
         std::priority_queue<std::pair<dist_t, labeltype >> topResults;
-        if (cur_element_count == 0) return topResults;
-        for (int i = 0; i < k; i++) {
+        dist_t lastdist = std::numeric_limits<dist_t>::max();
+        for (int i = 0; i < cur_element_count; i++) {
             dist_t dist = fstdistfunc_(query_data, data_ + size_per_element_ * i, dist_func_param_);
-            labeltype label = *((labeltype*) (data_ + size_per_element_ * i + data_size_));
-            if ((!isIdAllowed) || (*isIdAllowed)(label)) {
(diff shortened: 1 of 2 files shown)


dyashuni:
@lukaszsmolinski Thank you for the PR!
Good catch. Looks good.

dyashuni:
For some reason, this code
```
std::priority_queue<std::pair<dist_t, hnswlib::labeltype >> result = alg->searchKnn(
    (void*)items.data(row), k, p_idFilter);
for (int i = k - 1; i >= 0; i--) {
    auto& result_tuple = result.top();
    data_numpy_d[row * k + i] = result_tuple.first;
    data_numpy_l[row * k + i] = result_tuple.second;
    result.pop();
}
```
from [here](https://github.com/nmslib/hnswlib/blob/develop/python_bindings/bindings.cpp#L875-L881) works even if `result.size() < k`.

Could you please add an exception if `result.size() != k` to prevent unexpected results from being returned?
Something like this
```
if (result.size() != k)
  throw std::runtime_error(
      "Cannot return the results in a contiguous 2D array. Probably ef or M is too small");
```
from [here](https://github.com/nmslib/hnswlib/blob/develop/python_bindings/bindings.cpp#L646-L648)

lukaszsmolinski:
Done. By the way, it seems like this function is also missing the `if (normalize)` check. I can fix that in this PR if you'd like.

dyashuni:
Thank you, @lukaszsmolinski !
Yes, please add query normalization to `BFIndex` search to align it with `Index`

dyashuni (approved):
Thank you!

PR #515, opened 2023-10-22
Repair graph method

dyashuni:
Sometimes we can get 0 in-degree nodes after updates. This adds a new method to find 0 in-degree nodes and add new connections to them.


chasingegg:
Sometimes we get 0 in-degree nodes without updates.. Graph connectivity can not be guaranteed.

PR #536, opened 2024-02-08
Update bindings.cpp

IbraheemTaha:
Add a python function **get_max_layer** which returns the **maxlevel_** of the index.


IbraheemTaha:
I added a python function **get_node_degree** binding to the C++ function **getNodeDegree** which returns the count of neighbors of a node given its label.

IbraheemTaha:
I added the Python function **get_average_degree_at_layer** and the corresponding C++ function **getAverageDegreeAtLayer0** which returns the average node degree of a given layer (level). 

These functions is to get more insights and internal information about the structure of the index.

Issue #547, opened 2024-03-19
CMakeLists needs a fix for some hardcoded flags


barracuda156:
Some of the hardcoded flags here are wrong: https://github.com/nmslib/hnswlib/blob/master/CMakeLists.txt

1. `-march=` is not supported on PowerPC, `-mtune=` is.

2. `-lrt` is a BSD library, it should not be used on macOS, AFAIK.

yurymalkov:
Hi @barracuda156,
Can you please fix it wit a PR (and maybe test)?

barracuda156:
@yurymalkov Will it run CI here? I will check on macOS locally, but I cannot check it on Linux or BSD.
https://github.com/nmslib/hnswlib/pull/550

PR #550, opened 2024-03-22
CMakeLists: fix GCC flags

barracuda156:
Fixes: https://github.com/nmslib/hnswlib/issues/547


CMakeLists.txt
```diff
@@ -48,7 +48,17 @@ if(HNSWLIB_EXAMPLES)
...
-      SET( CMAKE_CXX_FLAGS  "-Ofast -lrt -std=c++11 -DHAVE_CXX0X -march=native -fpic -w -fopenmp -ftree-vectorize -ftree-vectorizer-verbose=0" )
+      SET( CMAKE_CXX_FLAGS  "-Ofast -std=c++11 -DHAVE_CXX0X -fpic -w -fopenmp -ftree-vectorize -ftree-vectorizer-verbose=0" )
+      if(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc|powerpc")
+        SET( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mtune=native" )
+        message("set -mtune=native flag")
+      else()
+        SET( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native" )
+        message("set -march=native flag")
+      endif()
+      if(NOT APPLE)
```
dyashunin-curie, line 59: I think we check MacOS several lines above
`if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")`

Do we need to check it here again?

dyashuni, line 59: I think we check MacOS several lines above
`if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")`

Do we need to check it here again?

barracuda156: What is the related of these two?

Clang can be used on non-Apple systems, and GCC can be used on Apple.

dyashuni: Thank you for explanation

dyashuni:
Thank you, @barracuda156

CMakeLists.txt
barracuda156: Clang is used as the default compiler in Xcode since macOS 10.7, but GCC can be used with no issues on any macOS from 10.4 to 14.4.1.
Clang is also a default compiler in the recent versions of FreeBSD.
So by no means it should be seen as “Clang = Apple ”.

PR #556, opened 2024-04-13
Declare saveIndex/loadIndex as templates to use custom streams

drons:


drons:
With this patch It is possible to implement load/store to any external storage or parallel index load (#524) at user level.

PR #576, opened 2024-07-11
support DataLevel0BlocksMemory data struct

Axlgrep:
support DataLevel0BlocksMemory data struct to solve the realloc huge memory doubling problem

1. The current data_level0_memory_ continuous memory block design can result in memory doubling during large index resize, which is very unfriendly to memory and performance(the realloc operation may trigger an expensive data copy action)，so we redesigned data_level0_memory_ (using multiple small memory blocks linked together).
2. In addition, we found some obvious _mm_prefetch out-of-bounds access issues and fixed them (there may be other hidden issues, please check them as well).

@yurymalkov Please take a look at this MR, thank you...

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ Full list of changes: https://github.com/nmslib/hnswlib/pull/523
 ### Highlights:
 1) Lightweight, header-only, no dependencies other than C++ 11
 2) Interfaces for C++, Python, external support for Java and R (https://github.com/jlmelville/rcpphnsw).
-3) Has full support for incremental index construction and updating the elements. Has support for element deletions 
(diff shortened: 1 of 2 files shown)


yurymalkov:
Thank you for the PR!

1) I wonder if you've tested the performance - that it does not affect search/build speed?
2) Can you add a few tests (also check backward compatibility)?
3) Can you update the branch? - the CI was broken

Axlgrep:
> Thank you for the PR!
> 
> 1. I wonder if you've tested the performance - that it does not affect search/build speed?
> 2. Can you add a few tests (also check backward compatibility)?
> 3. Can you update the branch? - the CI was broken

PR #619, opened 2025-03-26
Add support for disabling exceptions and improve CI

michaelbautin:
Adding support for disabling exceptions and improving CI

Adding a mode of error handling that does not rely on C++ exceptions,
with Abseil-like Status and StatusOr return types (but not adding a
dependency on Abseil). The old exceptions-based API is still available,
and the new functions are suffixed with "NoExceptions".

Refactoring the CMake build system. Organizing examples and tests and
reducing repetition. Adding the HNSWLIB_ENABLE_EXCEPTIONS option (ON by
default). Exceptions are also enabled when building Python bindings, and
in a few examples/tests where the code currently relies on exception
handling, even if HNSWLIB_ENABLE_EXCEPTIONS is set to OFF.

Adding ctest support. The only tests not supported by ctest are
test_updates invoked with and without the "update" argument, because
that requires installing numpy and running a data generation script and
doing that from a CMake test might be controversial. Those tests are
still run as part of GitHub Actions CI.

Enabling sanitizer support using ENABLE_ASAN / ENABLE_UBSAN /
ENABLE_TSAN / ENABLE_MSAN CMake options. Adding a combined ASAN/UBSAN
build type to the C++ test matrix in GitHub Actions in non-Windows
builds. Also adding Clang compiler on Ubuntu to the build matrix.

Making sure that assertions are enabled during tests. Using the
RelWithDebInfo build type and customizing it to remove the NDEBUG flag.
Also removing the default /EHsc flag when building with MSVC and instead
enabling exceptions when needed.

Fixing race conditions when accessing random number generators and
distributions from multiple threads in multiThreadLoad_test. These race
conditions were caught using ASAN/UBSAN.


dyashuni:
@michaelbautin Looks good!

I have some concerns regarding the use of `auto` in the C++ examples. For less experienced users, this might make the code harder to understand. Could you please keep the original types, even if they are a bit long?

michaelbautin:
@dyashuni @yurymalkov : thanks for the review and for validating the approach! I will clean up the code / tests, restore the explicit types, and update the PR.

michaelbautin:
@yurymalkov , @dyashuni : I've updated the PR, addressed the review comments, added ctest support and ran the tests locally. Could you please trigger CI/CD again on this PR?

michaelbautin:
@yurymalkov , @dyashuni : I think the PR is ready for both of you to take another look. The test matrix has been significantly enhanced and the tests are passing.

tests/cpp/multivector_search_test.cpp
dyashunin-curie, line 1: I think this file was accidentally modified, right?
After the change it tests resize.

tests/cpp/resize_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+    // Resize the index and add the second batch
+    std::cout << "Resizing the index to " << kNumElements << "." << std::endl;
+    alg_hnsw->resizeIndex(kNumElements);
+
+    const int second_batch_size = kNumElements - first_batch_size;
+    std::cout << "Adding the second batch of " << second_batch_size
+              << " elements." << std::endl;
+    for (int i = first_batch_size; i < kNumElements; ++i) {
+      alg_hnsw->addPoint(data.data() + (i * kDim), i);
+    }
```
dyashunin-curie, line 56: Lets add a check that after resize we can retrieve all points
```
    // Final validation - ensure all points are retrievable
    std::cout << "Final validation of all elements..." << std::endl;
    for (int i = 0; i < kNumElements; ++i) {
        auto result = alg_hnsw->searchKnn(data.data() + (i * kDim), 1);
        assert(!result.empty() && result.top().second == i);
    }
    
    std::cout << "Resize test completed successfully!" << std::endl;
```

hnswlib/hnswalg.h
```diff
@@ -1407,5 +1504,29 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
         std::cout << "integrity ok, checked " << connections_checked << " connections\n";
     }
+
+private:
+
+    /*
+    * Uses the last 16 bits of the memory for the linked list size to store the mark,
+    * whereas maxM0_ has to be limited to the lower 16 bits, however, still large enough in almost all cases.
+    */
+    Status markDeletedInternal(tableint internalId) {
```
dyashunin-curie, line 1494: I think we can move `unmarkDeletedInternal` to private as well. Let me discuss it with @yurymalkov

dyashunin-curie (changes requested):
@michaelbautin Looks very good, added a couple of refactoring comments

tests/cpp/resize_test.cpp
michaelbautin: Done.

tests/cpp/multivector_search_test.cpp
michaelbautin: Good catch, fixed.

hnswlib/hnswalg.h
dyashunin-curie: @yurymalkov says that it is fine to make `unmarkDeletedInternal` private as well. Lets do it.

dyashuni:
@michaelbautin Good job! Thank you!

michaelbautin:
Thanks @dyashuni @yurymalkov! Is there anything else to update, or could this be merged?

dyashuni:
We can merge @michaelbautin

PR #628, opened 2025-06-18
Fixing invalid prefetch pointers

michaelbautin:
This fix was originally authored by @yairgott.

Also enabling prefetch instructions in ASAN/UBSAN tests because they would not fail anymore with prefetch enabled with this fix.


dyashuni:
@michaelbautin this fix will decrease the speed. See https://github.com/nmslib/hnswlib/pull/459

PR #632, opened 2025-09-08
Add a function to get internal node id by user label

michaelbautin:
Add a thread-safe function getInternalIdByLabel to get internal node id by user label. Use that function in getDataByLabel.


dyashunin-curie (approved):
Thank you, @michaelbautin 
Looks good to me.

yurymalkov (approved):
Thank you!

PR #653, opened 2026-03-21
MakeLists: fix GCC flags (rebased #550)

ilyajob05:
Fixes: https://github.com/nmslib/hnswlib/issues/547
MakeLists: fix GCC flags (rebased #550)

Rebased version of #550 by @barracuda156 onto current develop.

Original changes:
- Fixed flags for PowerPC 

Resolved merge conflicts during rebase. Original authorship preserved.


CMakeLists.txt
```diff
@@ -162,10 +162,17 @@ if(HNSWLIB_EXAMPLES)
             endif()
         endif()
     elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-        # We enable optimizations in all build types, even in Debug.
-        add_cxx_flags(
-            -Ofast -lrt -march=native -w -fopenmp -ftree-vectorize
-            -ftree-vectorizer-verbose=0)
+      add_cxx_flags(-Ofast -std=c++11 -DHAVE_CXX0X -fpic -w -fopenmp -ftree-vectorize -ftree-vectorizer-verbose=0)
```
barracuda156: My earlier PR had those flags simply because the original code had them, and the PR only dealt with fixing one specific issue while minimizing changes made. Setting C++11 is not gcc-specific thing, and currently the flag is not added conditionally on GNU, which is correct, so we should not add it back here. The same probably applies to `-fpic` (better check that).

barracuda156 (changes requested):

barracuda156:
Thank you for resurrecting this. I think minor changes are due, but it will be good to have it merged once those are done.

CMakeLists.txt
ilyajob05: Thank you for the comment
I've fixed it https://github.com/nmslib/hnswlib/pull/653/changes/fd9f638cfe5827464eb506389d5e5d95f57b4782

dyashuni (approved):

PR #656, opened 2026-03-23
[Fix] Serialization / deserialization of enter factor...

kiplingw:
Serialize and deserialize the user's enter factor parameter.


yurymalkov:
That would break the compatibility between saves done with different versions.

kiplingw:
True. One way to fix that, albeit not retroactively, is to write out a binary version and check it when reading back. Thoughts?

yurymalkov:
I think if the change is made it needs to be backwards compatible (e.g. an old checkpoint without version would still be loaded).

kiplingw:
It's difficult to reliably determine older data since the first data written out is a `size_t` which could have a value of anything.

PR #531, opened 2023-12-21
remove unused variable

lulyon:
std::streampos position is never used.

PR #658, opened 2026-03-25
Fix bruteforce results, update CI, and version bump to 0.9.0

ilyajob05:
**version 0.9.0**

* Fixed incorrect results in bruteforce search with filter (#514) by [@lukaszsmolinski](https://github.com/lukaszsmolinski)
* Fixed missing normalization check in BFIndex (#514) by [@lukaszsmolinski](https://github.com/lukaszsmolinski)
* Throw an exception when fewer than k elements are available (#514) by [@lukaszsmolinski](https://github.com/lukaszsmolinski)
* Remove unused variable (#531) by [@lulyon](https://github.com/lulyon)
* Change cosine similarity to distance in README by [@yurymalkov](https://github.com/yurymalkov)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,11 +4,14 @@ on: [push, pull_request]
 
 jobs:
   test_python:
-    runs-on: ${{matrix.os}}
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
+        exclude:
+          - os: macos-latest
+            python-version: "3.7"
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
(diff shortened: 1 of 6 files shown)


kiplingw:
@ilyajob05, if I can make a suggestion, I'd centralize the version into a standalone textfile the build environment parses at configuration time since only making it available in the Python bindings means there's no versioning in the underlying C++ header library.

ilyajob05:
> @ilyajob05, if I can make a suggestion, I'd centralize the version into a standalone textfile the build environment parses at configuration time since only making it available in the Python bindings means there's no versioning in the underlying C++ header library.

@kiplingw  Thanks for the suggestion! I agree that centralizing the version would be a nice improvement 😎
I suggest we consider this as a separate merge request to the develop branch, which will be next in line for the release version

kiplingw:
I think that is a good idea. After a versioning PR is merged, I should then update #602 so that pkg-config has that versioning information available.

yurymalkov (approved):
Thank you!

dyashuni (approved):
Looks good, thank you!

PR #661, opened 2026-03-27
fix pointer type to guarantee correct alignment

hunglongtrangithub:
The external label (`labeltype`, aka `size_t`) is stored in a packed byte buffer (`data_level0_memory_` in `HierarchicalNSW`, `data_` in `BruteforceSearch`). Its address within the buffer is not guaranteed to be 8-byte aligned, so reads must go through a stack-allocated `labeltype` variable via `memcpy`. The pointer into the buffer should therefore be typed as `char*`, not `labeltype*`.

This silences UBSan warnings and avoids potential mis-compilation from the compiler assuming the pointer is aligned.

### DIFF
--- a/hnswlib/bruteforce.h
+++ b/hnswlib/bruteforce.h
@@ -4,6 +4,7 @@
 #include <mutex>
 #include <algorithm>
 #include <assert.h>
+#include "hnswlib.h"
 
 namespace hnswlib {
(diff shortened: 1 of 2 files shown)


hunglongtrangithub:
Will open a new PR to the develop branch

PR #683, opened 2026-09-15
Refactor no-exceptions load/save after v0.10.0-rc.2 review

ilyajob05:
- Drop `StreamExceptionsOff` / `invokeWithoutStreamThrow`. `*NoExceptions` I/O already returns `Status` via `!good()` / `peek`. Callers must leave the default iostream mask (`goodbit`)
- Fail-closed load: do not `clear()` a live HNSW index until the on-disk layout checks out. `clear()` resets lookup / deleted set / `max_elements_`. Reload no longer keeps stale labels or `deleted_elements`
- `BruteforceSearch::loadIndexNoExceptions`; throwing `loadIndex` is a wrapper. Rebuild the label map; memcpy labels; reject `size_per_element` mismatch vs `Space`
- `Status` copies the message (no dangling stack `const char*`)
- CMake: do not `FORCE`-strip `/EHsc` from the parent cache. `HNSWLIB_ENABLE_EXCEPTIONS` is example/test-only; MSVC OFF uses `/EHs-c-` on the target


hnswlib/bruteforce.h
```diff
@@ -169,32 +168,86 @@ class BruteforceSearch : public AlgorithmInterface<dist_t> {
...
-        readBinaryPOD(input, size_per_element_);
-        readBinaryPOD(input, cur_element_count);
 
-        data_size_ = s->get_data_size();
+        size_t data_size = s->get_data_size();
+        size_t size_per_element = data_size + sizeof(labeltype);
+        if (file_size_per_element != size_per_element) {
+            return Status("Cannot load index: size_per_element does not match space");
+        }
+        char *new_data = (char *) malloc(file_maxelements * size_per_element);
```
michaelbautin, line 193: It would be preferable to use RAII to manage this allocation.

hnswlib/bruteforce.h
```diff
@@ -169,32 +168,86 @@ class BruteforceSearch : public AlgorithmInterface<dist_t> {
...
+        if (new_data == nullptr)
+            return Status("Not enough memory: loadIndex failed to allocate data");
+        input.read(new_data, file_maxelements * size_per_element);
+        if (!input) {
+            free(new_data);
+            return Status("Cannot load index: failed to read vector data");
+        }
+
+        std::unordered_map<labeltype, size_t> new_dict;
+#if defined(__EXCEPTIONS) || _HAS_EXCEPTIONS == 1
```
michaelbautin, line 203: Instead of repeating `defined(__EXCEPTIONS) || _HAS_EXCEPTIONS == 1` in multiple places, we could define a `HNSWLIB_...` macro, e.g.

In hnswlib.h:
```c++ 
#if defined(__EXCEPTIONS) || _HAS_EXCEPTIONS == 1
#define HNSWLIB_EXCEPTIONS_ENABLED 1
#else
#undef HNSWLIB_EXCEPTIONS_ENABLED
#endif
```

Then here
```c++
#ifdef HNSWLIB_EXCEPTIONS_ENABLED
try {
#endif
```
etc.

michaelbautin (changes requested):

